PR #12, opened 2013-02-03
Added the work we already did.

bheesham:
In @spratt 's repo was the work we already did.
I'd like for us to merge it, then work in branches.

For now, this is going to be our working copy, until we rewrite most of the code.


bheesham:
Thoughts anyone?

spratt:
Looks fine, though I was going to get it to the point where we can run a copy locally with sqlite.  I'll keep working on that after merging this much.

PR #14, opened 2013-02-04
Easy Deploy (no DB)

spratt:
- Lots of backend cleanup.
- Added install instructions.
- Added simple deploy scripts.


bheesham:
These look good.

How many people do we need to agree before we merge? Or are we doing a time based-merged?

spratt:
One person is enough.  Just as long as the code doesn't have anything clearly bad on a first reading.

PR #17, opened 2013-02-04
Developer Deployment (with DB!!)

### LINKED ISSUE #13: Easy Developer Deployment
A requirement is for it to be as easy as possible for a new member like @psimonyi or a TA like @jbeltram to pull down the repo and run our code.

Since I have experience deploying the code on our production server and several dev servers, I'll take lead in making this happen.

spratt:
Closes #13 

Deploying in dev mode (using bin/start_dev.sh) creates a sqlite database (which will be ignored by git), and runs the server using that as a persistent store.

Also fixed:
- syntax highlighting (broken because I used a symlink in production)
- commenting (broken for the same reason)


cdelahousse:
Yay value!

PR #18, opened 2013-02-05
Added var directory

spratt:
After trying to run the scripts on Bhee's netbook, we found that they
rely on the existence of a var directory to store server logs and pids.


spratt:
Further addresses #13

spratt:
I'm going to reject this and take Bhee's instead, since he did his commit first.

PR #44, opened 2013-02-05
Made sure that Simon's scripts work. They do on my linux machine.

bheesham:
Was needed to get the setup script working that @spratt created.

```
new file:   var/.folder
```


spratt:
Further addresses #13

PR #54, opened 2013-02-15
Modelines

psimonyi:
Since we have literal tab characters that aren't supposed to be 8 wide, can we have modelines?


cdelahousse:
Simon isn't vim user. I'm not sure about Bhee.

Modelines are a pretty small convenience to be littering the top of all our files.

bheesham:
I'll be using vim to do this assignment, so I don't mind.

bheesham:
@spratt is the only one who uses emacs, so it's up to him? I'm sure emacs has something equivalent.

psimonyi:
Emacs has modelines too.

Alternatively, we could just use spaces for indentation...

spratt:
Emacs has some kind of modeline.  We should use them.  When we start minimizing the js, it ought to strip it out anyway.  Also, we should be using spaces.  Tabs are the worst, since the standard is they're worth 8 spaces, but different editors/viewers show them differently.

cdelahousse:
If Bhee's OK with them, the I HAVE to be OK with them. Sure, let's have modelines.

One request: could we have them at the bottom of files. I find them unpleasing when they're near the top.

psimonyi:
Do they work at the bottom? I thought they had to be within the first 5 lines.

spratt:
Also, the tabs are probably my fault.  By default, emacs mixes spaces and tabs.  It knows tab characters are worth 8 spaces, but the default indent width is 4 spaces.  So it will add 4 spaces for a single indent, a tab character for a double indent, a tab and 4 spaces for triple indent.....  Of course, I didn't know this when I first started writing the codebase.

bheesham:
I'd rather them at the top? It can go into the block comments.

spratt:
Looks fine to me.

PR #55, opened 2013-02-18
Code Cleanup

### LINKED ISSUE #35: Code clean up
_Imported issue by **[spratt](http://github.com/spratt)** from 2012-07-05T13:57:30Z_
_Originally opened as https://github.com/SteamedPears/CodeReviewClientMaster/issues/22_

---

The code is starting to get a little hairy.  We should probably do some serious review and clean it up.  There's lots of code duplication, both within view.js and duplicated code across index.js and view.js.  We should refactor to use the javascript module pattern, and separate the code into shared code and index/view specific code.

spratt:
Cleaned up ALL the code!

closes #35 

I broke the code into model modules:
- code
- comment
- language

View modules:
- editor
- view

And controller modules:
- main

I updated the version of CodeMirror we're using, and it's worth looking at whether this new version addresses the problems we saw.  It seems a lot faster, though it's still finicky sometimes.

We're now using requireJS to load the modules.

I ditched the feature to combine diffs for simplicity, though the code to do this is still in the repo if someone wants to pull it back in.

Bonus feature: the app doesn't reload to switch from the code submit form (what I call code mode) and code view / comment submit form (what I call comment mode).


spratt:
u mad??

rewrite

cdelahousse:
Apparently, this can't be merged automatically. Since this is your baby, @spratt, would you like to merge it?

spratt:
k.

PR #59, opened 2013-02-21
Fixed case sensitive issues

spratt:
Also fixed formatting in client-side scripts (sorry).


spratt:
I think this actually ruined the formatting on client-side js...I hate tab spaces...argh.

Issue #58, opened 2013-02-19
 /do/newcode -> Header: Access-Control-Allow-Origin: *


cdelahousse:
I need this http header so that the bookmarklet can submit stuff to CR.

spratt:
Cool.  I'm assigning this to you.

spratt:
Closed by pull request #63

PR #63, opened 2013-02-22
API CORS headers on response and /do/newcode now supports adding JSON requests

cdelahousse:
Formidable would crash when I'd send it JSON
Also, for my bookmarklet, I need a special header on the JSON response.


spratt:
closes #58

PR #65, opened 2013-02-22
Added a setup script.

bheesham:


bin/setup_dev.sh
```diff
@@ -0,0 +1,69 @@
...
+
+######################################################################
+# Configuration
+
+NODE_VERSION="0.8.20"
+
+# The directory where the node.tar.gz file will be downloaded and
+# extracted
+HOME_DIR="CHANGEME"
+EXTRACT_NODE="CHANGEME"
```
spratt, line 19: This will almost certainly result in someone changing this script then pushing their changed version.  I deliberately tried to avoid this when writing the original scripts.

bin/.start_dev.sh
```diff
@@ -16,7 +16,8 @@ SERVER_DIR="src/server"
 DB_INFO_FILE="models/db_info.js"
```
spratt, line 16: Wait.  Why did this change name to .start_dev.sh ?

bin/.start_dev.sh
```diff
@@ -16,7 +16,8 @@ SERVER_DIR="src/server"
...
 PID_FILE="var/server.pid"
 LOG_FILE="var/server.log"
-NODE_EXE="node"
+NODE_EXE=REPLACE_NODE_EXE
+NPM_EXE=REPLACE_NPM_EXE
```
spratt, line 20: This will also get changed and pushed.  This is what the path variable is for.

bheesham:
Guess I could do export PATH=$PATH:/path/to/node/bin/dir

bin/setup_dev.sh
psimonyi, line 19: Why not just use `$HOME`? Doesn't cygwin set that for you? (If not, `export HOME=${HOME:-$HOMEDRIVE$HOMEPATH}`)

bin/setup_dev.sh
```diff
@@ -0,0 +1,69 @@
...
+
+echo "Removing unneccessary files"
+rm -f "$EXTRACT_NODE/node.tar.gz"
+mv "$EXTRACT_NODE/node-v$NODE_VERSION-linux-$SYSTEM_ARCH" "$EXTRACT_NODE/node"
+
+NODE_EXE=$EXTRACT_NODE/node/bin/node
+NPM_EXE=$EXTRACT_NODE/node/bin/npm
+
+echo "Creating npm configuration, using all default values"
+$EXTRACT_NODE/node/bin/npm config ls -l > $HOME_DIR/.npmrc
```
psimonyi, line 60: You just finished setting `$NPM_EXE` to `$EXTRACT_NODE/node/bin/npm`, so you might as well use it.

bheesham:
Will make changes and stuff after.

Issue #57, opened 2013-02-19
Stop tracking other peoples' code


spratt:
Most of the code in the repo is library code on the client-side.  We should really use something so we stop tracking other peoples' code.

cdelahousse:
Agreed! I've been learning about them here.

http://git-scm.com/book/en/Git-Tools-Submodules

They sound swell.

psimonyi:
I don't like tracking other people's code, but git modules seem less than ideal. There seem to be a lot of pitfalls, and it seems to be intended for submodules maintained by the same people that own the main repo, rather than for using releases of other people's code. If there's no better way, then let's go for it, but if there's an alternative that still avoids tracking others' code, I'd like to hear it.

spratt:
We discussed this and determined the proper course of action is to pull down and build the projects we need in the install scripts.

spratt:
So I would like to have our install script pull down the sources for jQuery, CodeMirror, URI.js, require.js, and any other libraries we end up using on the client-side.  Does anyone know of a nice system for doing this?  Or should I roll a basic package manager using wget?  @cdelahousse @bheesham @psimonyi

cdelahousse:
Look into Yeoman, or more specifically, Bower (http://twitter.github.com/bower/). It's supposed to be a package manager for web apps.

If not, why not just do a series of git clones?

spratt:
In the cases of CodeMirror and URI.js, git cloning would work.  But we don't use the result of simply cloning the jquery and require.js repositories, we use some kind of built product.

psimonyi:
Since this should all be handled by a script, we don't need a nice package-manager interface -- all that package management would get us is dependency resolution (not really useful) and automatic building (potentially useful). So if you can find a package manager that will do building stuff automatically, let's try it; otherwise, wget is probably less work and less code. In the case of jQuery, it's just one file to download.

spratt:
The package manager also gives us that if the URL changes, it's the package manager's responsibility to know about it, we don't have to update our scripts.

psimonyi:
In the case of bower, it seems you give it URLs except for the really popular stuff, so that's probably not really an advantage for bower. But yes, if there's a package manager that does what we need, let's use it.

PR #73, opened 2013-03-04
Bookmarklet

cdelahousse:
Here's a working version of the bookmarklet.


src/bookmarklet/Makefile
```diff
@@ -0,0 +1,34 @@
+
+.PHONY=main
+
+TARGET=../client/
+
+main:
+	python -m SimpleHTTPServer
```
spratt: This is a new requirement and should be documented as such.

src/bookmarklet/Makefile
```diff
@@ -0,0 +1,34 @@
...
+
+lint:
+	jshint bookmarklet.js
```
spratt, line 13: We should probably start using jshint for most of our code.

src/bookmarklet/Makefile
```diff
@@ -0,0 +1,34 @@
...
+	cp vendors/* $(TARGET)vendors/
+
+clean:
+	rm -rf *.min.js $(TARGET)bookmarklet.js $(TARGET)bookmarklet.min.js \
+		$(TARGET)vendors/
+
+production:
+	cat bookmarklet.js |\
+		sed 's/\(g\.debug\s*=\s*\)true/\1false/' |\
+		uglify
```
spratt: Another new requirement that should be documented.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "src/bookmarklet/lib/readabilitySAX"]
+	path = src/bookmarklet/lib/readabilitySAX
+	url = git@github.com:fb55/readabilitySAX.git
```
psimonyi: The resolution in #57 was to pull in libraries as part of the "build", instead of git submodules. Talk to @spratt about what to do here.

src/bookmarklet/Makefile
```diff
@@ -0,0 +1,34 @@
...
+	cat ./vendors_src/readability_header.js \
+			./lib/readabilitySAX/readabilitySAX.js \
+			./lib/readabilitySAX/browsers/DOMasSAX.js \
+			./vendors_src/readability_footer.js \
+			> ./vendors/readability.js
+
+build:
+	@echo "Building project to $(TARGET)"
+	cp bookmarklet.js $(TARGET)
+	mkdir -p $(TARGET)vendors/
```
psimonyi: I think it would be clearer (and more traditional) to have `TARGET=../client` and then `mkdir -p $TARGET/vendors`. Then it doesn't matter whether `TARGET` ends with a slash or not.

src/bookmarklet/bookmarklet.js
```diff
@@ -0,0 +1,155 @@
+(function(win,doc,undefined) {
```
psimonyi: Do we really want to let Alexis' paranoia infect our code in general? Actually, this seems worse than trusting that `undefined` is not redefined, because it provides a way to redefine `undefined` without making it look as if that's what the code does.

src/bookmarklet/bookmarklet.js
```diff
@@ -0,0 +1,155 @@
...
+
+var g = {}; //Globals
+g.debug = true; //Debug mode
+g.vendors = {
+  gist : /https?:\/\/gist.github.com/  //Github gists
+};
+g.vendorDefault = "readability";
+g.baseUrl = g.debug ? '//localhost:8080' : '//review.steamedpears.com';
+g.bookmarkletUri = g.debug ? '/' : '/bookmarklet/';
+g.vendorUri = g.debug ? '/vendors/': '/bookmarklet/vendors/';
```
psimonyi: Space before `:`, please.

src/bookmarklet/vendors_src/readability_footer.js
```diff
@@ -0,0 +1,16 @@
...
+		var readable = new Readability();
+		readable.setSkipLevel(3); //XXX
+		saxParser(document.childNodes[document.childNodes.length-1], readable);
+		return readable.getArticle("text");
+	}
+}
+
+})(window,document);
+
+/* To wrapped around ../lib/readabilitySAX.js and ../lib/browsers/DOMasSax.js*/
```
psimonyi: I don't understand this comment. It doesn't seem to be a sentence.

src/bookmarklet/index.html
```diff
@@ -0,0 +1,28 @@
...
+
+
+	<a href="javascript:(function(){window.codeReview&&window.codeReview.exitApp();window.codeReview_script=document.createElement('script');codeReview_script.src='//localhost:8000/bookmarklet.js';(document.head||document.body).appendChild(codeReview_script)})();">Local</a>
+
+<script> 
+	(function(){
+		window.codeReview && window.codeReview.exitApp(); //delete prior codeReview instance
+		window.codeReview_script = document.createElement('script');
+		codeReview_script.src ='//localhost:8000/bookmarklet.js';
+		(document.head || document.body).appendChild(codeReview_script);
```
psimonyi, line 22: When will document.body exist but document.head not?

src/bookmarklet/bookmarklet.js
```diff
@@ -0,0 +1,155 @@
...
+   *  var tag =  createTag("http://example.com/someJSfile", "js");
+   */
+  var createTag = function createTag(url, t) {
+    var tag, type;
+
+    var matches = url.match(ext_re);
+    type = t || matches &&  matches[0].substr(1) || "default";
+
+    switch (type) {
+      case 'css' :
```
psimonyi: Extra space before the colon...

src/bookmarklet/bookmarklet.js
```diff
@@ -0,0 +1,155 @@
...
+        tag = doc.createElement('link');
+        tag.type= 'text/css';
+        tag.rel='stylesheet';
+        tag.href=url;
+        break;
+
+      case 'js':
+      case 'javascript':
```
psimonyi: I'd either put both cases on the same line, or add a `// fall through` comment to the `case 'js':` line.

src/bookmarklet/bookmarklet.js
```diff
@@ -0,0 +1,155 @@
...
+    */
+    runOnLoad  : function (asset, callback) {
+      if (asset.readyState){  //IE
+        asset.onreadystatechange = function(){
+          if (asset.readyState == "loaded" ||asset.readyState == "complete"){
+            asset.onreadystatechange = null;
+            return callback();
+          }
+        };
+      } else {  //W3C
```
psimonyi: Could you make the standard way the default? if (asset.addEventListener) { w3c way} else {try IE way}. That way, the standard code gets used if IE supports both.

src/bookmarklet/bookmarklet.js
```diff
@@ -0,0 +1,155 @@
...
+  };
+})();
+
+var helpers = {
+  //Figure out if this url uses a known vendor
+  vendorGetFromUrl : function(url) {
+    var matches,
+    regex;
+
+    for (var vendor in g.vendors) {
```
psimonyi, line 136: This probably needs a `hasOwnProperty` test. Maybe we should have a utility function to make it easier. Anyway, some library could extend `Object.prototype` and mess this up.

.gitmodules
psimonyi: The interface for using submodules seems more onerous than having a script pull in the dependencies.

src/bookmarklet/bookmarklet.js
```diff
@@ -0,0 +1,155 @@
+(function(win,doc,undefined) {
+'use strict';
+
+var g = {}; //Globals
+g.debug = true; //Debug mode
+g.vendors = {
+  gist : /https?:\/\/gist.github.com/  //Github gists
+};
+g.vendorDefault = "readability";
+g.baseUrl = g.debug ? '//localhost:8080' : '//review.steamedpears.com';
```
psimonyi, line 22: I'm not sure I like hard-coding our URL, but I don't know what else to suggest.

src/bookmarklet/bookmarklet.js
psimonyi: Couldn't you just host the bookmarklet in /bookmarklet on the local server too? I'd rather keep the difference between debug mode and normal mode small.

src/bookmarklet/bookmarklet.js
```diff
@@ -0,0 +1,155 @@
...
+    var xhr = new XMLHttpRequest();
+    xhr.open(method || 'POST', url, true);
+    xhr.setRequestHeader("Content-type", type || "application/json");
+    return xhr;
+  }
+};
+
+//Expose API
+that.dbg = dbg;
+window.codeReview = that;
```
psimonyi: This should probably be `win`, since it's a parameter you don't use anywhere else. Or you could rename the parameter to `window`. That might be better.

src/bookmarklet/bookmarklet.js
```diff
@@ -0,0 +1,155 @@
...
+  var assets = {},
+      head = doc.getElementsByTagName('head')[0];
+
+  var ext_re= /\.[0-9a-zA-Z]+$/; // matches a URL's extension
+  
+  /* 
+   * Create tag that will be appended to head
+   * Returns an HTML element
+   * Usage:
+   *  var tag = createTag("http://example.com/style.css");
```
psimonyi: This will create a `<link>` tag with no `rel` attribute -- is that ever a good thing?

src/bookmarklet/bookmarklet.js
```diff
@@ -0,0 +1,155 @@
...
+
+      case 'js':
+      case 'javascript':
+        tag = doc.createElement('script');
+        tag.src = url;
+        break;
+
+      default:
+        tag = doc.createElement('link');
+        tag.href=url;
```
psimonyi: There are a lot of places where you're inconsistent with spaces; I'm going to stop pointing them out individually.

src/server/requestHandlers.js
```diff
@@ -129,6 +130,21 @@ function newcode(request,response) {
...
+	if (request.method && request.method.toUpperCase() === 'OPTIONS') {
+		response.writeHead(
+				"204",
+				"No Content",
+				{
+					"access-control-allow-origin": "*",
+					"access-control-allow-methods": "GET, POST, PUT, DELETE, OPTIONS",
+					"access-control-allow-headers": "content-type, accept",
+					"access-control-max-age": 10, // Seconds.
+					"content-length": 0
```
psimonyi: It would probably be best to use the canonical capitalization of these names: `Content-Length` etc.

src/server/requestHandlers.js
```diff
@@ -129,6 +130,21 @@ function newcode(request,response) {
 				return error(response,500,'Error writing code to database');
 			});
 	}
+	if (request.method && request.method.toUpperCase() === 'OPTIONS') {
+		response.writeHead(
+				"204",
+				"No Content",
+				{
+					"access-control-allow-origin": "*",
+					"access-control-allow-methods": "GET, POST, PUT, DELETE, OPTIONS",
```
psimonyi: Do we do anything for PUT and DELETE?

psimonyi: If not, it seems odd to allow them here.

src/bookmarklet/vendors_src/readability_header.js
```diff
@@ -0,0 +1,9 @@
+/* To wrapped around ../lib/readabilitySAX.js and ../lib/browsers/DOMasSax.js*/
+(function(window,document,undefined) {
+'use strict';
+
+if (!window.codeReview)
+	throw new Error("codeReview object not loaded.");
```
psimonyi: It's easy to forget to add braces when you add another consequent line, so I'd prefer either putting this all on one line or adding braces. I'll bring this up in the style guide.

src/bookmarklet/Makefile
psimonyi, line 13: Also, this is a new requirement.

src/server/requestHandlers.js
bheesham: I don't ever see us using DELETE for the bookmarklet, but I can see us using PUT.

src/bookmarklet/bookmarklet.js
bheesham: I don't think you need it. [Or so w3c doesn't say](http://www.w3.org/TR/REC-html40/struct/links.html#edef-LINK).
Bad practice: probably, bad code: probably not.

src/bookmarklet/bookmarklet.js
bheesham, line 22: Go through their history and find a website with our title in it? I think this is one of those places where hard-coding is required, unless there's a way to find out information from X without ever specifying what or where is X.

psimonyi, line 22: I was thinking more along the lines of having the URL set at compile time.

src/bookmarklet/bookmarklet.js
psimonyi: I don't think we should be putting bad practices in the documentation showing how to use something. And the HTML5 spec says "A link element must have either a rel attribute or an itemprop attribute, but not both." --[WHATWG](http://www.whatwg.org/specs/web-apps/current-work/multipage/semantics.html#the-link-element).

src/server/requestHandlers.js
psimonyi: Okay, but it's easy to change when we start supporting those things. Changing the bookmarklet will be a lot harder than changing the server code.

src/bookmarklet/Makefile
cdelahousse: Addressed.
0a85f21e11aa6dd2ee3f533743c9de61ef13aebf

src/bookmarklet/Makefile
cdelahousse, line 13: See documentation.
0a85f21e11aa6dd2ee3f533743c9de61ef13aebf

src/bookmarklet/Makefile
cdelahousse: 0a85f21e11aa6dd2ee3f533743c9de61ef13aebf

src/bookmarklet/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+To prepare the bookmarklet for production (eg. minification, concatenation, etc), use `make production`.
+
+##Dependencies
+
+* Python for a static file server. Node's static file server caches static files. Using python's HTTP server is a work around.
+* uglifyjs for minifying.
+* jshint for linting.
+
+All dependencies are optional. They're only used in the makefile. Use npm's `-g` flag to install uglify and jshint as terminal commands.
```
psimonyi: These lines must be more than 80 characters long...

src/bookmarklet/README.md
```diff
@@ -0,0 +1,17 @@
+#Code Review Bookmarklet
+
+##Development Usage
+
+Run `make main` to start serving the  bookmarklet for development.
+
+All vendors must be built before they can be used. Use `make vendorname`.
```
psimonyi: It would help to introduce what a 'vendor' is.

src/bookmarklet/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+To prepare the bookmarklet for production (eg. minification, concatenation, etc), use `make production`.
+
+##Dependencies
+
+* Python for a static file server. Node's static file server caches static files. Using python's HTTP server is a work around.
```
psimonyi: What node module were you using?

src/bookmarklet/README.md
cdelahousse: 5a4545a278231bb0546cdf0d24d3d991636f7f7d

src/bookmarklet/README.md
cdelahousse: 5a4545a278231bb0546cdf0d24d3d991636f7f7d

src/bookmarklet/README.md
cdelahousse: 5a4545a278231bb0546cdf0d24d3d991636f7f7d
node-static

src/bookmarklet/README.md
```diff
@@ -0,0 +1,39 @@
...
+* [ReadabilitySAX](https://github.com/fb55/readabilitySAX) is a readability
+  fork that extracts the textual from a page using a system of heuristics.
+
+Vendor source files are located in `libs/`.
+
+##Build Dependencies
+
+* Python for a static file server. `node-static` caches static files so
+  reloading the browser didn't always fetch the newest version of the
+  bookmarklet. Using python's HTTP server is a work around.
```
psimonyi: According to the [docs](https://github.com/cloudhead/node-static#cache), you can configure that.

.gitmodules
cdelahousse: bff578b

src/bookmarklet/README.md
cdelahousse: I don't want to change the node server just for developing the bookmarklet. The python web server is simple solution.

psimonyi: I don't understand what you're serving -- it seems to be just the compiled bookmarklet, so I don't understand why you even need a server. But if you do need a server, since caching is a parameter, it won't affect other node servers.

src/bookmarklet/bookmarklet.js
cdelahousse: This "undefined" redefinition is a common pattern in JS. I see it a lot.  Check out the first line of code on jQuery:
http://code.jquery.com/jquery-1.9.1.js

I will not be changing this.

src/bookmarklet/bookmarklet.js
cdelahousse: e8917a63a00d79e4c4bf77029078fc25f22c50ad

src/bookmarklet/bookmarklet.js
cdelahousse: Whitespace: e8917a63a00d79e4c4bf77029078fc25f22c50ad

About your second comment: yes. I'll get to it.

src/bookmarklet/bookmarklet.js
cdelahousse: 12c1463

src/bookmarklet/Makefile
cdelahousse, line 13: Linted the code:
e8917a63a00d79e4c4bf77029078fc25f22c50ad

src/bookmarklet/bookmarklet.js
cdelahousse, line 22: Really good idea, @psimonyi. I've added it to my todo list.

src/bookmarklet/index.html
cdelahousse, line 22: It was an IE7 or 6 bug. I read about it a long time ago, but forgot the specifics.

If a head doesn't exist, I can add a script tag to the body. Both are valid ways of adding a script tag to a document.

src/bookmarklet/bookmarklet.js
cdelahousse: Changed win to window and doc to document:
d6386c6

src/bookmarklet/bookmarklet.js
cdelahousse: d6386c679dfdc402e93456fe9448bda2c967ffd9

src/bookmarklet/bookmarklet.js
spratt, line 22: Should we address this in a future commit?

src/server/requestHandlers.js
spratt: Right now and in the foreseeable future we'll only be using the POST method with CORS.

src/bookmarklet/bookmarklet.js
psimonyi, line 22: Yes. I think this is fine for now.

src/server/requestHandlers.js
psimonyi: I think it makes most sense to permit only those methods we actually use. And this only applies to /do/newcode anyway, which should only ever get POST and OPTIONS, right? So `POST, OPTIONS` is sufficient.

src/server/requestHandlers.js
```diff
@@ -129,6 +130,21 @@ function newcode(request,response) {
 				return error(response,500,'Error writing code to database');
 			});
 	}
```
psimonyi, line 153: There are four cases this function handles: OPTIONS, POST with form data, POST with JSON, and anything else (error case). That seems like a reasonable order. But only the POST cases use the `writetodb` function, so I'd put the new OPTIONS code before that.

src/bookmarklet/lib/.gitignore
```diff
@@ -0,0 +1,4 @@
+# Ignore everything in this directory
+*
+# Except this file
+!.gitignore
```
psimonyi: Since `.gitignore` is already tracked, there's no point in excluding it from the gitignore. (Remember, gitignore only applies to adding untracked files.)

src/bookmarklet/bookmarklet.js
```diff
@@ -0,0 +1,159 @@
...
+g.newCodeApi = '/do/newcode';
+
+
+//Will be the exposed API
+var that, 
+    codeReview;
+
+codeReview = that = {};
+
+function dbg() { if (g.debug) console.log.apply(console,arguments); }
```
psimonyi, line 24: Since you've cleverly kept all debugging stuff to single lines, we could strip these lines at compile time... (Future goal.)

src/bookmarklet/bookmarklet.js
psimonyi: On Tue, 2013-03-05 at 09:05 -0800, Christian Delahousse wrote:

> In src/bookmarklet/bookmarklet.js:
> 
> > +(function(win,doc,undefined) {
> > This is a common pattern in JS. I see it a lot. Check out the first
> > line of code on jQuery:
> > http://code.jquery.com/jquery-1.9.1.js
> 
> I will not be changing this.

Okay. I won't object to merging code with this pattern. But I don't like
it. I haven't seen it in any JS code except Alexis's demo code (and now
jQuery). Furthermore, I think it's pretty much useless clutter. Since
ES5, the `undefined` property of the global object is not writable, so
the only way for the identifier `undefined` to have a value other than
the undefined value is for a lexically enclosing context to contain a
`var` statement defining `undefined` (or to contain a suitably twisted
`with` statement, but that's prohibited in strict mode). And even if
your paranoia extends to ES3, I think `var undefined;` is a better way
to be sure `undefined` means the undefined value, because it isn't
spread across two widely separated lines and it doesn't _look_ as if
you're doing something dumb with `undefined`.

src/bookmarklet/vendors_src/readability_footer.js
cdelahousse: 3a6b9f3

src/bookmarklet/README.md
cdelahousse: db725a6aa32ec7fbcecbacac95e1c7923a0ee2be

Node-static is now serving the file. The python dependency has been removed.

src/server/requestHandlers.js
cdelahousse: ab94169

Only using Post and Options.

src/server/requestHandlers.js
cdelahousse: 9793387

src/bookmarklet/bookmarklet.js
cdelahousse, line 136: cf41111

src/server/requestHandlers.js
cdelahousse, line 153: a56e991

src/bookmarklet/lib/.gitignore
cdelahousse: It doesn't hurt. I'm not changing this.

src/bookmarklet/bookmarklet.js
spratt: This still needs to be addressed I think?

cdelahousse: 32353f1

Removed the last case and make the default case be JS

src/bookmarklet/Makefile
```diff
@@ -0,0 +1,54 @@
...
+
+readability:
+	@echo "Wrapping $@ and compiling to $(VENDORS)"
+	cat $(VENDORS_SRC)/readability_header.js \
+			$(LIB)/readabilitySAX/readabilitySAX.js \
+			$(LIB)/readabilitySAX/browsers/DOMasSAX.js \
+			$(VENDORS_SRC)/readability_footer.js \
+			> $(VENDORS)/readability.js
+
+getvendors:
```
psimonyi, line 25: Either make this .PHONY or name the rule after the folder it creates.

src/bookmarklet/Makefile
```diff
@@ -0,0 +1,54 @@
...
+	@echo "Downloading vendors"
+	cd $(LIB); git clone git@github.com:fb55/readabilitySAX.git
+
+buildvendors: readability
```
psimonyi, line 30: .PHONY

src/bookmarklet/Makefile
```diff
@@ -0,0 +1,54 @@
...
+	mkdir -p $(TARGET)
+	mkdir -p $(TARGET)/$(VENDORS)
+
+#Build the bookmarklet, compile/wrap the vendors, move everything to the
+#TARGET directory
+build: makeproductiondirs buildvendors
+	cp bookmarklet.js $(TARGET)
+	cp vendors/* $(TARGET)/$(VENDORS)
+
+clean:
```
psimonyi, line 50: .PHONY

src/bookmarklet/Makefile
```diff
@@ -0,0 +1,54 @@
+.PHONY=main
```
psimonyi: There are several other PHONY rules.

src/bookmarklet/bookmarklet.js
```diff
@@ -0,0 +1,166 @@
...
+
+var assetLoader = (function() {
+  var assets = {},
+      head = document.getElementsByTagName('head')[0];
```
psimonyi: This could be `document.head`.

src/bookmarklet/bookmarklet.js
```diff
@@ -0,0 +1,166 @@
...
+   * Returns an HTML element
+   *
+   * Default type: javascript
+   *
+   * Usage:
+   *  var tag = createTag("http://example.com/style.css");
+   *  var tag = createTag("http://example.com/someJSfile", "js");
+   *
+   */
+  var ext_re= /\.[0-9a-zA-Z]+$/; // matches a URL's extension
```
psimonyi: Since this is only used inside `createTag`, it should be declared inside too. (Also, space before `=` while you're at it.)

src/bookmarklet/bookmarklet.js
```diff
@@ -0,0 +1,166 @@
...
+      if (g.vendors.hasOwnProperty(vendor)) {
+        regex = g.vendors[vendor];
+        matches = regex.test(url);
+        if (matches) return vendor;
+      }
+    }
+    return null;
+  },
+  //Load a given vendor
+  vendorLoad: function (vendor, callback) {
```
psimonyi: `vendorLoad` sounds like an event handler. `loadVendor`?

src/bookmarklet/bookmarklet.js
```diff
@@ -0,0 +1,166 @@
...
+    dbg("helpers.vendorLoad: " + vendor);
+    assetLoader.load(g.baseUrl + g.vendorUri + vendor + ".js", callback);
+  }, 
+  //Cors Request
+  Request : function (method, url, type){
```
psimonyi: Is there a reason this starts with a capital? (Also: no space before `:`)

src/bookmarklet/vendors/.gitignore
```diff
@@ -0,0 +1,5 @@
+# Ignore everything in this directory
+*
+# Except this file
+!.gitignore
+*.js
```
psimonyi: You already ignore `*`, so ignoring `*.js` is redundant.

src/bookmarklet/vendors_src/readability_footer.js
```diff
@@ -0,0 +1,16 @@
+
+dbg("readability.js: Readability loaded");
+
+window.codeReview.vendor = {
+	getContent : function () {
+		var readable = new Readability();
+		readable.setSkipLevel(3); //XXX
```
psimonyi: What's the XXX about? When someone comes back to look at this, what do they need to think about?

src/server/requestHandlers.js
```diff
@@ -110,6 +110,28 @@ function comments(request,response) {
 ******************************************************************************/
 function newcode(request,response) {
 
+	/****************************************************************************
+	*	Preflight OPTIONS requests -- CORS																				*
```
spratt: TABS ARE GROSS

src/server/requestHandlers.js
```diff
@@ -110,6 +110,28 @@ function comments(request,response) {
...
+					"Access-Control-Allow-Headers": "content-type, accept",
+					"Access-Control-Max-Age": 10, // Seconds.
+					"Content-Length": 0
+				});
+
+		return response.end();
+	}
+
+	/****************************************************************************
+	* POST requests																															*
```
spratt: TABS NOOOOOOO

spratt:
I am signing off on the server code.

src/bookmarklet/vendors/.gitignore
cdelahousse: 744777e

src/bookmarklet/Makefile
psimonyi: Also, this isn't the right syntax. `.PHONY: target-name`
And I like the style where you put that immediately before the rule being made phony.

src/bookmarklet/bookmarklet.js
```diff
@@ -0,0 +1,166 @@
...
+g.vendorDefault = "readability";
+g.baseUrl = g.debug ? '//localhost:8080' : '//review.steamedpears.com';
+g.bookmarkletUri =  '/bookmarklet/';
+g.vendorUri = g.debug ? '/vendors/' : '/bookmarklet/vendors/';
+g.newCodeApi = '/do/newcode';
+
+
+//Will be the exposed API
+var that, 
+    codeReview;
```
psimonyi: `codeReview` is defined but never used.

src/server/requestHandlers.js
cdelahousse: ed982041cc94982f

src/server/requestHandlers.js
cdelahousse: ed982041cc94982f

src/bookmarklet/vendors_src/readability_footer.js
cdelahousse: 40027f3cf6eb314fab67

src/bookmarklet/bookmarklet.js
```diff
@@ -0,0 +1,196 @@
...
+
+var g = {}; //Globals
+g.debug = true; //Debug mode
+g.vendors = {
+  gist: /https?:\/\/gist.github.com/  //Github gists
+};
+g.vendorDefault = 'readability';
+g.baseUrl = g.debug ? '//localhost:8080' : '//review.steamedpears.com';
+g.bookmarkletUri =  '/bookmarklet';
+g.vendorUri = g.debug ? '/bookmarklet/vendors/' : '/bookmarklet/vendors/';
```
psimonyi: If (cond) then x, else x.

src/client/.gitignore
```diff
@@ -4,3 +4,5 @@ do/
 
 # Removed the shortcut
 include/CodeMirror
+
+bookmarklet/*
```
psimonyi, line 8: Didn't you say this was no longer needed?

cdelahousse, line 8: Bookmarklet.js isn't needed but the bookmarklet will be built to
client/bookmarklet/

src/bookmarklet/bookmarklet.js
```diff
@@ -0,0 +1,196 @@
...
+        }
+      }
+    }
+  };
+})();
+
+/*******************************************************************************
+* Helpers                                                                      *
+*******************************************************************************/
+var helpers = {
```
psimonyi, line 130: Not sure what the benefit of putting these in such a generic object is. [non-blocking]

src/bookmarklet/index.html
```diff
@@ -0,0 +1,28 @@
...
+  <h1>Code Review bookmarklet test page</h1>
+
+
+	<p>Drag the following link to you bookmarks bar<\p>
+	<a href="javascript:(function(){window.codeReview&&window.codeReview.exit();window.codeReview_script=document.createElement('script');codeReview_script.src='//localhost:8080/bookmarklet.js';(document.head||document.body).appendChild(codeReview_script)})();">Local</a>
+
+<script> 
+	(function(){
+		window.codeReview && window.codeReview.exit(); //delete prior codeReview instance
+		window.codeReview_script = document.createElement('script');
```
psimonyi: Why is codeReview_script a global variable?

src/bookmarklet/index.html
```diff
@@ -0,0 +1,28 @@
...
+<section id="main-content" role="main">
+  <h1>Code Review bookmarklet test page</h1>
+
+
+	<p>Drag the following link to you bookmarks bar<\p>
+	<a href="javascript:(function(){window.codeReview&&window.codeReview.exit();window.codeReview_script=document.createElement('script');codeReview_script.src='//localhost:8080/bookmarklet.js';(document.head||document.body).appendChild(codeReview_script)})();">Local</a>
+
+<script> 
+	(function(){
+		window.codeReview && window.codeReview.exit(); //delete prior codeReview instance
```
psimonyi, line 19: I would much prefer `if (window.codeReview) window.codeReview.exit();`. This way feels too much like Perl.

psimonyi, line 19: Also, `if (codeReview) ...` is sufficient.

src/bookmarklet/Makefile
```diff
@@ -0,0 +1,61 @@
...
+VENDORS_SRC=vendors_src
+VENDORS=vendors
+
+.PHONY: lint
+lint:
+	jshint bookmarklet.js
+
+#Initialize the bookmarklet
+.PHONY: init
+init: makedirs getvendors build
```
psimonyi: `init` should probably be the default rule (i.e. be the first rule in the file) so you can just run `make`. [blocking]

src/bookmarklet/bookmarklet.js
cdelahousse: 1a429b2

psimonyi:
Looks good; I'm just trying to run it. Is `index.html` being served?

src/bookmarklet/index.html
cdelahousse: Because it needs to persist and codeReview hasn't been loaded yet.

src/bookmarklet/index.html
cdelahousse, line 19: I used a completely legit JavaScript convention. I'm not changing this.

cdelahousse:
No. Index.html is only used have a link to drag to your bookmarks bar.  It's just a bit of test code that will disappear soon.

src/bookmarklet/index.html
psimonyi: Why does it need to persist? I can't see where you use it again.

psimonyi:
Oh, okay. I'll just look at the code then.

src/bookmarklet/Makefile
cdelahousse: 7009289

src/bookmarklet/index.html
psimonyi, line 19: I think this is really gross, but won't block on it.

src/bookmarklet/index.html
cdelahousse: 3ea4bae

psimonyi:
Looks good.

PR #77, opened 2013-03-08
Stop tracking client library files, fixes #57

### LINKED ISSUE #57: Stop tracking other peoples' code

spratt:


bin/install_client.sh
```diff
@@ -0,0 +1,69 @@
...
+# Install dependencies
+
+cd $LIB_DIR
+
+for f in $RAW_DEPS; do
+	filename=`echo $f | sed -e 's!.*/!!'`
+	if [ -f $filename ]; then
+		continue
+	fi
+	wget $f
```
spratt: Undocumented new dependency: wget

[blocking]

spratt: 766f28b

bin/reset_client.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+######################################################################
+# Configuration
+
+CLIENT_DIR="src/client"
+LIB_DIR="scripts/lib/"
```
psimonyi: I think it would be clearer to have `LIB_DIR="$CLIENT_DIR/scripts/lib`.
Is there a reason for the difference in trailing slashes?

spratt: No reason.

spratt: 9114871

bin/install_client.sh
```diff
@@ -0,0 +1,69 @@
...
+	echo "$RAW_DEPS_FILE not found"
+	exit 2
+fi
+
+if [ ! -f $ZIP_DEPS_FILE ]; then
+	echo "$ZIP_DEPS_FILE not found"
+	exit 3
+fi
+
+RAW_DEPS=`cat $RAW_DEPS_FILE`
```
psimonyi: `$(< $RAW_DEPS_FILE)` maybe? [non-blocking]

bin/start_dev.sh
```diff
@@ -20,6 +20,15 @@ NODE_EXE="node"
 BUILD_DB_SCRIPT="models/build_db.js"
 TEST_DATA_SCRIPT="models/test_data.js"
 INDEX_SCRIPT="index.js"
+CLIENT_INSTALL_SCRIPT="bin/install_client.sh"
+
+######################################################################
+# Install Client-Side
+$CLIENT_INSTALL_SCRIPT
```
psimonyi: I don't think this needs to be a variable...

bin/reset_client.sh
```diff
@@ -0,0 +1,20 @@
...
+if [ ! -d $LIB_DIR ]; then
+	echo "Please run this file from the root directory of the project"
+	exit 1
+fi
+
+######################################################################
+# remove all files in the lib directory
+
+cd $LIB_DIR
+rm -rf *
```
psimonyi: Or just `rm -rf $LIB_DIR/*` [non-blocking]
Are there ever any dotfiles? (If `$LIB_DIR/.*` matches anything, they won't be deleted.)

psimonyi:
I'm signing off on these changes.

bin/install_client.sh
```diff
@@ -0,0 +1,58 @@
...
+	download $f
+	exitIfFailed "Failed to get $f"
+done
+
+for f in `cat "$ZIP_DEPS_FILE"`; do
+	download $f
+	exitIfFailed "Failed to get $f"
+	unzip -quo `basename $f`
+	exitIfFailed "There was an error unzipping `basename $f`"
+done
```
cdelahousse, line 58: This code is beautiful.

bin/start_dev.sh
```diff
@@ -22,6 +22,14 @@ TEST_DATA_SCRIPT="models/test_data.js"
 INDEX_SCRIPT="index.js"
 
 ######################################################################
+# Install Client-Side
+bin/install_client.sh
```
cdelahousse, line 26: Do we want to run this install script every time we use the start script. Can't this just be done at the discretion of the developer?

bin/install_client.sh
spratt, line 58: Aww, shucks.  :+1:

src/client/raw.deps.dev.txt
```diff
@@ -0,0 +1,10 @@
+http://files.simondavidpratt.com/js/URI.js
+http://files.simondavidpratt.com/js/IPv6.js
+http://files.simondavidpratt.com/js/punycode.js
+http://files.simondavidpratt.com/js/SecondLevelDomains.js
+http://files.simondavidpratt.com/js/diff_match_patch.js
```
cdelahousse, line 5: Are you going to be in charge of keeping these files on your server up to date? Can't we just point to the raw files within their repos?

bin/start_dev.sh
spratt, line 26: bin/install_client.sh really just makes sure all the libraries are installed.  If they exist, it does nothing.

src/client/raw.deps.dev.txt
spratt, line 5: We shouldn't be "keeping them up to date" per se.  We should keep them at a stable version and update them when we update what version we're using.

Pointing at the raw files doesn't consistently work.  My wget fails because it doesn't trust the certificates.

cdelahousse:
Yeah. It's cool. My complaints are minor. I'll pull it in.

PR #89, opened 2013-03-20
Basic Tutorial

### LINKED ISSUE #24: Tutorial
_Imported issue by **[spratt](http://github.com/spratt)** from 2012-07-15T04:51:10Z_
_Originally opened as https://github.com/SteamedPears/CodeReviewClientMaster/issues/60_

---

Now that the interface is actually usable, we should start thinking about user experience.  When the user first submits code, they are presented with a page that shows their code and whole bunch of empty space.  If no comments have been made, we should have something which lets the user know they can select lines in the code to comment.

spratt:
closes #24 

Added
- emacs local variables
- bootstrap
- basic tutorial


src/client/index.html
```diff
@@ -5,18 +5,17 @@
...
     <div id="error"></div>
-    <div id="code_instructions">
-    Insert the code you would like reviewed then
-    hit submit, and share the url with your reviewers.
-    </div>
-    <div id="comment_instructions">
-      Review the code, select the lines of code on which you would like
-      to comment, then write a comment.
+    <div id="tutorial">
+      <a id="tutorialBtn" class="btn btn-large btn-success" href="javascript:void(0);">
```
psimonyi: We said we'd use dashes in IDs: `tutorial-btn`. Also, does this have to be an `a`, or could it be a `button`? I think a `button` would be better.

src/client/index.html
```diff
@@ -35,24 +34,29 @@
...
-      <input type="hidden" name="code_id" id="code-id" value="-1" />
-      <input type="hidden" name="line_start" id="line-start" value="-1" />
-      <input type="hidden" name="line_end" id="line-end" value="-1" />
-      <input type="text" name="user" id="user" value="Anonymous" />
-      <p>
-        Commenting on:
-        <span id="line-start-num">?</span>-<span id="line-end-num">?</span>
+      <div id="comment-new">
+        <form id="comment-form" method="POST" action="/do/newcomment">
+          <input type="hidden" name="code_id" id="code-id" value="-1" />
```
psimonyi: If you're going to reindent this, you might as well strip off the trailing `/` too.

src/client/scripts/tutorial.js
```diff
@@ -0,0 +1,78 @@
...
+  function codeMode() {
+    $(document).ready(function() {
+      // set up the code mode of the tutorial
+      tutorialBtn.click(function() {
+        console.log('starting code tutorial');
+
+        removeAllAttributes();
+
+        codeArea.attr('data-step','1');
+        codeArea.attr('data-intro','Enter the text you would like to be reviewed');
```
psimonyi: Over 80.

src/client/scripts/tutorial.js
```diff
@@ -0,0 +1,78 @@
...
+
+  function removeAllAttributes() {
+    codeArea.removeAttr('data-intro');
+    commentArea.removeAttr('data-intro');
+    codeArea.removeAttr('data-step');
+    commentArea.removeAttr('data-step');
+  }
+
+  function codeMode() {
+    $(document).ready(function() {
```
psimonyi, line 43: [non-blocking] It would be nice if we could guarantee that the DOM is loaded at an earlier point. If we wrapped main.js (or view.js?) in $(document).ready, we wouldn't have to do so anywhere else.

psimonyi, line 43: -- especially since the jQuery API makes no guarantees about the order in which these functions will be executed.

spratt, line 43: That's a really good point.  I think we can do a little better - the DOM stuff should be encapsulated within view.js and things it calls.  We could wrap the entry into that module and that ought to do it while allowing things like models to be loaded first.

src/client/scripts/tutorial.js
```diff
@@ -0,0 +1,78 @@
...
+      });
+    });
+  }
+  tutorial.codeMode = codeMode;
+
+  function commentMode() {
+    $(document).ready(function() {
+      // set up the comment mode of the tutorial
+      tutorialBtn.click(function() {
+        console.log('starting comment tutorial');
```
psimonyi: We should have a way to avoid shipping with debugging output. This could be a follow-up, though.

src/client/scripts/tutorial.js
```diff
@@ -0,0 +1,78 @@
...
+        removeAllAttributes();
+
+        codeArea.attr('data-step','1');
+        codeArea.attr('data-intro','Enter the text you would like to be reviewed');
+        
+        intro().start();
+      });
+    });
+  }
+  tutorial.codeMode = codeMode;
```
psimonyi: Why this way instead of `tutorial.codeMode = function () {`?

src/client/index.html
```diff
@@ -29,30 +26,35 @@
...
+            <input id="comment-submit-button" type="submit" value="Submit Comment">
+          </p>
+        </form>
+      </div>
     </div>
   </body>
 </html>
 
+<!-- Local Variables:         -->
+<!-- tab-width: 8             -->
```
cdelahousse: These lines are gross. Can you not get the Local Variables all on one line?

src/client/scripts/tutorial.js
spratt: I overheard some discussion about how giving names to functions makes them easier to debug.

src/client/scripts/tutorial.js
cdelahousse: What about abstracting it with a function, say dbg('starting comment tutorial')

src/client/scripts/main.js
```diff
@@ -20,7 +20,8 @@ require.config({
     'CodeMirror':'lib/codemirror-3.1/lib/codemirror',
     'jquery.form':'lib/jquery.form',
     'diff':'lib/diff_match_patch',
-    'jquery':'lib/jquery-1.9.1.min'
+    'jquery':'lib/jquery-1.9.1.min',
+    'intro':'lib/intro.js-0.1.0/minified/intro.min'
```
cdelahousse, line 24: Need space after colons.

src/client/scripts/main.js
```diff
@@ -46,6 +47,9 @@ require.config({
       init:function() {
         return new diff_match_patch();
       }
+    },
+    'intro':{
```
cdelahousse, line 51: Space after colon.

src/client/scripts/main.js
psimonyi, line 24: Better to match the existing style here, since he isn't reindenting everything anyway. It would be good to fix this in a separate commit.

src/client/scripts/view.js
```diff
@@ -11,10 +11,12 @@
 define([
   "jquery",
   "editor",
```
cdelahousse, line 13: Use single quotes, as per style guide.

psimonyi, line 13: Diffs look better if changes are confined. It would be better to fix the code he's not touching in a separate commit. We've got to get on that linting thing.

src/client/scripts/main.js
psimonyi, line 24: Actually, Christian's right. Perhaps it wouldn't be a bad idea to follow the style guide for the lines you touch (but leave the context lines alone, so the diff is still useful). Thoughts?

src/client/scripts/tutorial.js
psimonyi: You can still give the function a name: `tutorial.codeMode = function codeMode() {` is equivalent to what you've written. My feeling is that it looks better, but I'm pretty ambivalent about this.
Also, it turns out that modern debugging tools aren't too shabby with anonymous functions. Named functions are still nicer, but in many places they can find the name of the property to which the anonymous function was assigned.

src/client/scripts/tutorial.js
psimonyi: Yeah, that would work. I think the best outcome would be something that lets us strip them at build time (i.e. when we minify/concatenate/whatever). Some function would be a good start. A good logging module?

src/client/scripts/tutorial.js
```diff
@@ -0,0 +1,78 @@
...
+  "intro"
+], function($,intro) {
+  var tutorial = {};
+
+  // save some DOM elements
+  var tutorialBtn, codeArea, commentArea;
+
+  // initialization
+  $(document).ready(function() {
+    tutorialBtn = $("#tutorial-btn");
```
psimonyi: `"` → `'`

src/client/index.html
psimonyi: Yeah, it would be a lot better at 3 lines long. But if you're looking into directory-variables, I think we can merge this in the meantime. At least it's at the end.

spratt: No, it can't be one line, and it can't be 3 lines.  You're going to have to deal with it until I switch to directory variables. [wontfix]

src/client/scripts/main.js
spratt, line 24: If we're going to go back and fix the other lines anyway, I'd rather be consistent for now. [wontfix]

src/client/scripts/main.js
spratt, line 51: If we're going to go back and fix the other lines anyway, I'd rather be consistent for now. [wontfix]

src/client/scripts/view.js
spratt, line 13: If we're going to go back and fix the other lines anyway, I'd rather be consistent for now. [wontfix]

src/client/index.html
spratt: If we were allowing this at the top it could be 1 line.

spratt:
Fixed all outstanding issues that don't break consistency.

spratt:
Fixed emacs local variables issues by adding directory variables.

psimonyi:
I tried this locally, and I get `TypeError: n is null @ http://localhost:8080/scripts/lib/intro.js-0.1.0/minified/intro.min.js:1`. It never gets to part 2 of the comment-mode tutorial. Is this perhaps because the code-mode click handler is still active?

spratt:
I was having a different issue with the comment mode tutorial too.  When you don't select text on the code side, the comment tutorial can't highlight the new comment box because it's hidden.  Could this be the same error?

psimonyi:
I did select some text, so the comment box was visible. But I somehow
ended up with two layers of grey overlay.

spratt:
Try now.  The second part of the comment tutorial doesn't get added to the tutorial until you actually select some text.

This causes a weird thing where if you follow the tutorial, you will only get the first part (because when you started the tutorial there was no selected text).  But I think it works.

psimonyi:
As you say, it's weird if you try to follow the tutorial, but at least the really buggy bit is gone. We can fix that stuff in some follow-ups, right?

PR #90, opened 2013-03-20
Support per-language CodeMirror file dependencies (fixes #36)

### LINKED ISSUE #36: Github Flavored Markdown broken
_Imported issue by **[spratt](http://github.com/spratt)** from 2012-07-05T13:54:04Z_
_Originally opened as https://github.com/SteamedPears/CodeReviewClientMaster/issues/21_

---

This is caused by the fact that GFM may contain any other language, so it has to load ALL the languages.  We should implement some kind of progress bar, or any kind of indication that the page is still loading.

Update February 19th, 2013: after the front-end rewrite (and CodeMirror update), GFM is completely broken.

psimonyi:
When a language defines a "depends" list, require the files it lists there, too.
This makes "Github Flavored Markdown" work again.


psimonyi:
I'm not especially happy with the name "depends". The difference between "depends" and "requires" is that "requires" is a list of other languages, resolved recursively; "depends" is a list of other files in the CodeMirror distribution [without the .js ending, because requirejs is finicky about that] that are immediately required and presumably handle their own dependencies.

spratt:
We may want to revisit depends/requires later, but it looks good for now.

Issue #100, opened 2013-03-22
The proxy server in our code won't be used in production


spratt:
We should probably get rid of it.

Issue #99, opened 2013-03-22
The static server in our code won't be used in production


spratt:
We should probably ditch it.  @psimonyi is looking into a simple alternative.

Issue #92, opened 2013-03-22
Code Review does not work with Node v.0.10.0


cdelahousse:
I just upgraded my version of node version 0.10.0-2chl1 of nodejs via this PPA: https://launchpad.net/~chris-lea/+archive/node.js/. 

It broke CR. 

I had to revert to version 0.6.19 of Nodejs to get it working again.

I suspect this has to do with either the reverse proxy or node-static. I could connect via `wget` to our app on port 20193 and get a response, but when I tried port 8080 (the proxy) and 8888 (node-static), a connection was made but no response was returned. I was left hanging...

cdelahousse:
Seems related to this:
https://github.com/cloudhead/node-static/pull/86

PR #101, opened 2013-03-23
Converted back-end to use connect.js, fixes #98

### LINKED ISSUE #98: Our routing is a dirty hack
I wrote the simplest possible routing while learning to use node when I first wrote the app and it hasn't been touched since.  We may want to replace it with a micro-framework like https://github.com/gjritter/nerve/

Thoughts?

spratt:
YAY WE USE CONNECT NOW WE'RE COOL AND HIP

But seriously, it got rid of a lot of code.


spratt:
@cdelahousse I'm trying to test the bookmarklet, and I'm finding its build instructions to be impenetrable.  It gives me 3 make targets and doesn't really help me figure out which one (or combination of several?) I need to do.  The buildvendor, for example, suggests that it's important to do this before any vendors can be used?  Do I need to run that before build?  Or does that get run by build?  Does install call build?  Why is git giving me an error if I run make build then make install?  Am I not supposed to do that?  Why isn't the script doing any kind of check?  I'm scared and confused.

Also, how do I use the bookmarklet?  This should really be integrated into our user interface in such a way that users can actually use it.

spratt:
Note that this
- closes #92 since it gets rid of the node-static dependency
- closes #99 since it only loads dev_mode (including the static server code) if running in dev mode
- closes #100 same as previous (proxy code is included in dev_mode)

bin/start_dev.sh
```diff
@@ -58,9 +60,16 @@ if [ ! -f $DB_INFO_FILE ]; then
...
+
+######################################################################
 # Server initialization
 
 echo Starting development server
-NODE_ENV=development $NODE_EXE $INDEX_SCRIPT \
-    1>> $ROOT_DIR/$LOG_FILE 2>> $ROOT_DIR/$LOG_FILE &
+NODE_ENV=development $NODE_EXE $INDEX_SCRIPT &> $ROOT_DIR/$LOG_FILE &
 echo $! > $ROOT_DIR/$PID_FILE
+date +%s > $ROOT_DIR/$TIMESTAMP_FILE
```
psimonyi: `date --iso-8601=seconds` would give more readable filenames.

spratt: That code does not work on my box.

spratt: Would you prefer something like: `date "+%Y/%m/%d/%H:%M"`?

spratt: I guess forward slashes are a dumb idea, so maybe `date "+%Y.%m.%d.%H.%M"`?

psimonyi: On Sat, 2013-03-23 at 23:08 -0700, Simon David Pratt wrote:

> I guess forward slashes are a dumb idea, so maybe date "+%Y.%m.%d.%H.%M"?

Yeah, only ISO 8601 style is probably best:
+'%Y-%m-%dT%H:%M:%S%z'

spratt: Colons are a bad idea, they're the file delimiter on mac.

spratt: *Path delimiter.

psimonyi: They're path delimiters on *nix, but still okay in path names. However,
it's fine to have no delimiters:
+'%Y-%m-%dT%H%M%S%Z'

spratt: By path delimiter, I mean mac represents paths as usr:bin:foo

psimonyi: How do they use all the BSD and GNU stuff that assumes : is a path
delimiter (as in PATH=/bin:/usr/bin:$HOME/bin)? Anyway, the standard
allows you to omit the separators.

spratt: Yeah, they do some weird conversion magic so it looks like paths are separated by /.

src/server/index.js
```diff
@@ -1,71 +1,80 @@
...
- * Start the server                                                            *
- ******************************************************************************/
-try{
+* Start the server                                                            *
+******************************************************************************/
+var app = connect()
+  .use(connect.logger(devMode?'dev':'short'))
+  .use(connect.timeout(requestTimeout))
+  .use(function(request,response,next) {
+    // Preflight OPTIONS requests -- CORS
```
psimonyi: I think this means we'll allow cross-origin requests for everything, when we used to allow them only for `/do/newcode`, right?

spratt: Yup, totally forgot.

src/server/package.json
```diff
@@ -1,43 +1,38 @@
...
+    "connect":          ">=2.7.3",
+    "proxy-middleware": ">=0.3.1"
+  },
+  "repository": {
     "type": "git",
-    "url": "https://bitbucket.org/sdp/codereviewservermaster.git"
-  }
-  , "license": "ISC License"
-  , "engine" : ">=0.7.9-pre"
+    "url":  "https://bitbucket.org/sdp/codereviewservermaster.git"
```
psimonyi: This doesn't look current...

src/server/package.json
```diff
@@ -1,43 +1,38 @@
...
+  },
+  "license": "ISC License",
+  "engine" : ">=0.7.9-pre"
```
psimonyi, line 38: Is this accurate?

src/server/languageList.js
```diff
@@ -1,304 +0,0 @@
...
-    "description": "XQuery",
-  },
-  {
-    "id": 60,
-    "mode": "yaml",
-    "description": "YAML",
-  }
-];
-
-/* vim: set softtabstop=2 shiftwidth=2 tabstop=8 expandtab textwidth=80: */
```
psimonyi, line 304: I like the diffstats on this file!

src/server/package.json
spratt, line 38: Not a clue.  Want me to update it to the current version I'm using?

src/server/package.json
spratt: Yeah, this is our old repo.

src/server/languageList.js
spratt, line 304: Yeah, we weren't using it anyway....I think I just forgot to delete it a while back.

src/server/package.json
psimonyi, line 38: I guess it should be the minimum version any of us test with. (For me,
it's v0.10.1.) So sure, set it to whatever you're using.

spratt, line 38: Am also using 10.1.

bin/start_dev.sh
```diff
@@ -58,9 +61,16 @@ if [ ! -f $DB_INFO_FILE ]; then
 fi
 
 ######################################################################
+# Rotate logs
```
psimonyi: Nobody's developing this on Windows, right? So how about putting the log file in the right place and keeping a symlink to the latest one? As in, `node index.js > var/log/server-$TIME.log; ln -s -f var/log/server-$TIME.log var/server.log` (approximately).

spratt: If they're running the shell script, they would have to be using cygwin which I'm pretty sure supports symbolic links.  It feels a little weird to do this, but I think it's more elegant than saving a timestamp in an extra file.

.gitignore
```diff
@@ -15,3 +15,4 @@ npm-debug.log
 *.db
 *.DS_Store
 *~
+*.time
```
psimonyi: This is no longer needed.

spratt: Fixed in    ba5c966

bin/start_dev.sh
```diff
@@ -35,17 +38,15 @@ fi
...
 # Install needed packages
-
-ROOT_DIR=`pwd`
-
 cd $SERVER_DIR
-
-echo "Installing needed packages"
-npm install
+OUTDATED=`npm outdated 2> /dev/null`
+if [ ${#OUTDATED} -ne 0 ]; then
```
psimonyi: Instead of comparing the string length to 0, you should just check whether the string is nonempty.

```
if [ "$(npm outdated 2> /dev/null)" ]; then
```

is most concise. (You could use backticks instead of `$()`; I just think it looks odd inside quotation marks.)
Other options include `if [ "$OUTDATED" ]` and `if [ -n "$OUTDATED" ]` if you'd rather keep the variable.

spratt: Checking the length is much clearer than what seems to be using a string as a boolean.

psimonyi: Use the `-n` operator: `if [ -n "$(npm outdated 2> /dev/null)" ]; then`

And without `-n`, you're not using a string as a boolean; you're passing a single argument to the `test` builtin.

spratt: Agreed, so it's proper syntax for test but it's not syntax whose meaning is immediately obvious.  Also, I'm getting a substitution error when I try to evaluate that line within the brackets.

spratt: It looks like ${command} is not equivalent to `command` on my system.

spratt:
Oh man, I'm looking forward to pull request I have waiting to submit after this one.

bin/start_dev.sh
spratt: Oh herp.  Well, I fixed with backticks.  Good enough?

spratt:
I can, but the other one is a fork from these changes.  The diff would include all of these changes.

bin/start_dev.sh
spratt: I understand that from your previous comment.  I am saying that the pull request now includes your fix with backticks inside of quotes instead of the parens.  Want me to go back and replace the backticks with parens?

psimonyi: Nah, backticks are fine. We're using them all over anyway.

src/server/index.js
```diff
@@ -1,71 +1,70 @@
...
+];
 
 /******************************************************************************
- * Start the server                                                            *
- ******************************************************************************/
-try{
+* Start the server                                                            *
+******************************************************************************/
+var app = connect()
+  .use(connect.logger(devMode?'dev':'short'))
```
psimonyi: I'd prefer spaces around the `?` and `:` (but what I'd like most is to resolve the "Proposals" in the style guide).

spratt: I like all the proposals.  It's been the de facto style guide for some time now.  Let's consider it resolved.

spratt: I figured the jshint would catch those errors.  I guess I'll go back and check everything against the style guide.

Issue #84, opened 2013-03-15
Give /do/newcode/ the option of not returning a complete Code object 


cdelahousse:
We don't need the content sent back to us over the wire when add some to the DB. I want to create an option for the API not to send it.

An example of why this would be good is the bookmarklet. We only need confirmation and the uuid returned to the client.

spratt:
Actually, I think this is true for the web page too.  The web page already has all the code loaded.

cdelahousse:
Should the default be content returned or no content returned?

psimonyi:
Given that both of the current users don't need it, not sending it seems
like a good default.

spratt:
We may need to modify the client code a little, but the client has all the information it needs.

spratt:
I'm pushing this tentatively to Sprint 4.

spratt:
This is solved in the above pull request.  Turns out the front-end didn't need modification at all, it was ignoring everything except for the id it was getting back.

Issue #41, opened 2013-02-05
Don't grab all comments at once


cdelahousse:
_Imported issue by **[spratt](http://github.com/spratt)** from 2012-06-26T14:36:00Z_
_Originally opened as https://github.com/SteamedPears/CodeReviewClientMaster/issues/4_

---

We should probably just grab the comments in limited amounts.  50 at a time, maybe?  This should make the page more responsive when there are lots of comments.

cdelahousse:
_Imported comment by **[Gankro](http://github.com/Gankro)** from 2012-06-27T18:19:58Z_ *

---

As it's designed now we could store with the code a map of line numbers to comment numbers. Then only when someone clicks on the (#) do the comments get AJAX loaded in. I dunno how this would work with the back end, loading all the comments from one specific line, that is. 

If there's too many comments on one line we can just load in the first 50 or something and have a "more..." button at the bottom that loads the next 50. Again the main challenge here is a back-end matter.

cdelahousse:
_Imported comment by **[spratt](http://github.com/spratt)** from 2012-06-29T05:46:20Z_ *

---

On the backend, it would be easy just to get the comments on a specific line number.  Would it hurt responsiveness?  We may want to think about the case where there are hundreds of comments on one line.  Then again, we may want to leave that for our future selves to deal with.

cdelahousse:
_Imported comment by **[Gankro](http://github.com/Gankro)** from 2012-07-15T15:59:21Z_ *

---

Details on expected resolution strategy here: https://github.com/SteamedPears/CodeReviewClientMaster/issues/46#issuecomment-6987610

spratt:
Detailed resolution strategy in this repo: https://github.com/CarletonU-COMP2406-W2013/Steamed-Pears/issues/27

spratt:
The front end now loads comment counts from the comment model, then only loads comments on a particular line when the button is pressed.  All that's left is implementing this in the backend and setting the front end up to take advantage of it.

Issue #32, opened 2013-02-05
Security Audit


cdelahousse:
_Imported issue by **[spratt](http://github.com/spratt)** from 2012-07-06T16:00:31Z_
_Originally opened as https://github.com/SteamedPears/CodeReviewClientMaster/issues/33_

---

We really need to sit down and try to break the security before we demonstrate the app publicly.

spratt:
@cdelahousse @bheesham @psimonyi One or more of us should do this for the next (and final) sprint.  I want to get feedback from @asomayaji on how to secure our app before the end of the course.

spratt:
@psimonyi is going to be lead on this.  

We should meet with @asomayaji sometime this week to go over the common security flaws we could try to fix.

@cdelahousse @bheesham You both should come if you're interested in the security side.

spratt:
http://redis.io/topics/security for when we start using redis

cdelahousse:
Tell me you guys go see Anil. I'd love to see what he has to say.

cdelahousse:
From #102:
I think we should start using a replacer function argument with JSON.stringify.
https://developer.mozilla.org/en-US/docs/Using_native_JSON#The_replacer_parameter
@asomayaji was speaking about whitelists and a replace function would be a super easy implementation.

spratt:
@cdelahousse that's a great idea.

PR #102, opened 2013-03-25
Converted persistent store to use redis, closes #97 

### LINKED ISSUE #97: SQL or NoSQL?
I'm not sure we need the consistency guarantees of SQL.  Thoughts?

spratt:


src/client/scripts/comment.js
```diff
@@ -13,34 +13,23 @@ define([
 ], function($) {
   var comment = {};
 
-  var comments = null;
-
   comment.getCommentCounts = function(code_id,callback,error_fn) {
     comments = {};
-    $.ajax('/do/comments',{
+    $.ajax('/do/countComments',{
```
spratt: space after comma

src/client/scripts/comment.js
```diff
@@ -13,34 +13,23 @@ define([
...
-        callback(counts,comment.getCommentsOnLine);
-      }
+      success:  callback
     });
   };
 
-  comment.getCommentsOnLine = function(line) {
-    if(comments === null) return;
-    return comments[line];
+  comment.getCommentsOnLine = function(code_id,line,callback,error_fn) {
```
spratt: spaces after commas

src/client/scripts/comment.js
```diff
@@ -13,34 +13,23 @@ define([
...
+    $.ajax('/do/commentsOnLine',{
```
spratt: space after comma

src/client/scripts/comment.js
```diff
@@ -13,34 +13,23 @@ define([
...
+      data:   {code_id:code_id,line:line},
```
spratt: space after commas

src/client/scripts/main.js
```diff
@@ -75,11 +75,22 @@ require([
...
+      });
+    };
+  }
+
   // ajaxify forms
   $('#code-form').ajaxForm({
     success: function(ob) {
-      history.pushState({},"CodeReview","index.html?id="+ob.uuid);
-      view.initCommentMode(ob.uuid);
+      history.pushState({},"CodeReview","index.html?id="+ob.id);
```
spratt: spaces after commas

src/server/requestHandlers.js
```diff
@@ -36,6 +30,12 @@ function error(response, errno, errtext) {
   response.end();
 }
 
+function errorCheck(err,next) {
```
spratt: space after comma

src/server/requestHandlers.js
```diff
@@ -36,6 +30,12 @@ function error(response, errno, errtext) {
...
+  if(err !== null)
+    return error(response,500,'Error while writing to database.');
```
spratt: spaces after commas

src/server/requestHandlers.js
```diff
@@ -59,43 +59,67 @@ function isValidPositiveIntegerString(x) {
...
   }
-  Code
-    .find({where:{uuid:id}})
-    .success(function(code) {
-      if (code === null) {
-        return error(response, 404, 'Code not found');
-      }
-      return success(response, code);
-    });
+  client.get('code:'+id,function(err,reply) {
```
spratt: spaces after commas

src/server/requestHandlers.js
```diff
@@ -59,43 +59,67 @@ function isValidPositiveIntegerString(x) {
...
+    if(err !== null)
```
spratt: space after if

src/server/requestHandlers.js
```diff
@@ -59,43 +59,67 @@ function isValidPositiveIntegerString(x) {
...
+      return error(response,500,'Error while reading from database.');
+    if(reply === null)
```
spratt: space after if

src/server/requestHandlers.js
```diff
@@ -59,43 +59,67 @@ function isValidPositiveIntegerString(x) {
...
-    .find({where:{uuid:id}})
-    .success(function(code) {
-      if (code === null) {
-        return error(response, 404, 'Code not found');
-      }
-      return success(response, code);
-    });
+  client.get('code:'+id,function(err,reply) {
+    if(err !== null)
+      return error(response,500,'Error while reading from database.');
```
spratt: spaces after commas

src/server/requestHandlers.js
```diff
@@ -59,43 +59,67 @@ function isValidPositiveIntegerString(x) {
...
+    if(reply === null)
+      return error(response,404,'Code not found');
```
spratt: spaces after commas

src/server/requestHandlers.js
```diff
@@ -59,43 +59,67 @@ function isValidPositiveIntegerString(x) {
...
+    return success(response,JSON.parse(reply));
```
spratt: space after comma

src/server/requestHandlers.js
```diff
@@ -59,43 +59,67 @@ function isValidPositiveIntegerString(x) {
...
+  });
 };
 
-exports.comment = function comment(request, response) {
+exports.commentsOnLine = function commentsOnLine(request,response) {
```
spratt: space after comma

src/server/requestHandlers.js
```diff
@@ -59,43 +59,67 @@ function isValidPositiveIntegerString(x) {
...
   var query = url.parse(request.url, true).query;
-  var id = query.id;
-  if (id === undefined) {
-    return error(response, 400, 'Invalid comment id');
+  var id = query.code_id;
+  var line = query.line;
+  if(id === undefined) {
+    return error(response,400,'Invalid comment id');
```
spratt: spaces after commas

src/server/requestHandlers.js
```diff
@@ -59,43 +59,67 @@ function isValidPositiveIntegerString(x) {
...
-  Comment
-    .find(Number(id))
-    .success(function(comment) {
-      if (comment === null) {
-        return error(response, 404, 'Comment not found');
-      }
-      return success(response, comment);
-    });
+  if(line === undefined) {
+    return error(response,400,'Invalid line number');
```
spratt: spaces after commas

src/server/requestHandlers.js
```diff
@@ -59,43 +59,67 @@ function isValidPositiveIntegerString(x) {
...
+  }
+  client.lrange('comment:'+id+':'+line,0,-1,function(err,reply) {
```
spratt: spaces after commas

src/server/requestHandlers.js
```diff
@@ -59,43 +59,67 @@ function isValidPositiveIntegerString(x) {
...
+    if(err !== null)
```
spratt: space after if

src/server/requestHandlers.js
```diff
@@ -59,43 +59,67 @@ function isValidPositiveIntegerString(x) {
...
+      return error(response,500,'Error while reading from database.');
```
spratt: spaces after commas

src/server/requestHandlers.js
```diff
@@ -59,43 +59,67 @@ function isValidPositiveIntegerString(x) {
...
+    if(reply === null)
```
spratt: space after if

src/server/requestHandlers.js
```diff
@@ -59,43 +59,67 @@ function isValidPositiveIntegerString(x) {
...
+      return error(response,404,'Comment not found');
```
spratt: spaces after commas

src/server/requestHandlers.js
```diff
@@ -59,43 +59,67 @@ function isValidPositiveIntegerString(x) {
...
+    var out = [];
+    for(var i in reply) {
```
spratt: space after for

src/server/requestHandlers.js
```diff
@@ -59,43 +59,67 @@ function isValidPositiveIntegerString(x) {
...
+  });
 };
 
-exports.comments = function comments(request, response) {
+exports.countComments = function countComments(request, response) {
   var query = url.parse(request.url, true).query;
   var code_id = query.code_id;
-  Comment.findAll({where: {code_id: code_id}}).success(function(comments) {
-    return success(response, {code_id:code_id, comments:comments});
+  client.smembers('comment:'+code_id+':indices',function(err,reply) {
```
spratt: spaces after commas

spratt:
Note that this pull request
- closes #97 since it gives us a noSQL version to test at the very least
- closes #84 since submitting code now only returns an object with the id
- closes #41 since it splits /do/comments into /do/countComments which returns a map of line number to how many comments on that line, and /do/commentsOnLine which returns only the comments on a given line

INSTALL
```diff
@@ -11,7 +11,7 @@ Requirements:
  * node (v0.8.17, v0.8.20)
  * npm (1.2.11)
  * node-gyp (v0.8.4)
- * sqlite3 (3.7.15.1, 3.7.13)
+ * redis (2.6.10)
  * g++ (4.7.2)
```
spratt: I'm not sure g++ is a requirement any more, since we're not compiling sqlite.  We are compiling hiredis, but I think that requires gcc.  I'll look into this.

[blocking]

src/client/scripts/code.js
```diff
@@ -13,8 +13,8 @@ define([
 ], function($) {
   var code = {};
 
-  code.getCode = function(id,callback,error_fn) {
-    $.ajax('/do/code',{
+  code.getCode = function(id, callback, error_fn) {
```
cdelahousse, line 16: Will the client facing code/comment ID be the same as the Redis collection/key/hash for that entry? Isn't that a bit sketchy?

cdelahousse, line 16: OK, nevermind. I see what you're doing. The ID is just data member.

INSTALL
```diff
@@ -11,8 +11,8 @@ Requirements:
  * node (v0.8.17, v0.8.20)
  * npm (1.2.11)
  * node-gyp (v0.8.4)
- * sqlite3 (3.7.15.1, 3.7.13)
- * g++ (4.7.2)
+ * redis (2.6.10)
```
cdelahousse, line 14: Shouldn't this be redis-server instead of just redis? Do we want the package name or the project name? I'd kinda want to know what to `sudo apt-get install ...`.

cdelahousse:
I haven't run it, but everything looks solid. This is a major change, so I think others should go over it before we accept it.

INSTALL
spratt, line 14: The problem with packages is that they're system-dependent.  The apt package is named redis-server, but both the rpm and the mac port packages are called redis.  We have the same problem with node.  The name of the software is redis.

src/client/scripts/code.js
spratt, line 16: I can explain this further since it isn't clear: when code is submitted, we generate a random uuid for the code in node.  This is used to build the key in redis (of the form "code:uuid-key-goes-here") that refers to the JSON string representing the code object.

When a request like this comes in, the server-side requests the code with key "code:id-requested-by-ajax" which I think is fine?  But correct me if we should change this.

INSTALL
cdelahousse, line 14: I'm OK with this. 

Maybe we should have a line in our start scripts which detects the name of the appropriate binary?

spratt, line 14: That would be nice.  I'm sure there's a typical way to do this.  

A simple way to approximate this is to let the user pass in an alternative binary name.  We're using variables for the binaries any way, so we can detect whether the variable is already set, and if so - use that instead.  So the user can just run `NODE=node-app DB=redis-server bin/start_dev.sh`

src/client/scripts/comment.js
```diff
@@ -13,34 +13,23 @@ define([
...
+      success:  callback
     });
   };
 
-  comment.getCommentsOnLine = function(line) {
-    if(comments === null) return;
-    return comments[line];
+  comment.getCommentsOnLine = function(code_id, line, callback, error_fn) {
+    $.ajax('/do/commentsOnLine', {
+      data:   {code_id:code_id, line:line},
```
cdelahousse: I hate doing this, I'm not trying to piss you off, but, the style guide says there should be a space after a colon for methods.

src/client/scripts/main.js
```diff
@@ -65,7 +65,7 @@ require([
   "comment",
   "language",
   "editor"
-], function($,URI) {
+], function($, URI) {
   var view = require("view");
```
cdelahousse, line 69: Single quotes.

src/client/scripts/main.js
```diff
@@ -90,26 +101,26 @@ require([
...
-                               view.addCommentButtons,
+                               addButtons(ob.code_id),
                                view.displayError);
     }
   });
 
   // dispatch based on query
   var query = URI(document.URL).query(true);
-  if(query.error !== undefined)
+  if (query.error !== undefined)
```
cdelahousse: Wouldn't this be a bit clearer if the following statement would be on the same line or in curly braces. No big deal though.

src/server/requestHandlers.js
```diff
@@ -1,21 +1,14 @@
...
-var Code = sequelize.import(__dirname + '/models/code');
-var Comment = sequelize.import(__dirname + '/models/comment');
-
-// set the associations
-Code.hasMany(Comment, {as: 'Comments', foreignKey: 'code_id'});
+/******************************************************************************
+* Handle DB errors                                                            *
+******************************************************************************/
+redis.on('error', function (err) {
+  console.log('DB Error: ' + err);
```
cdelahousse: I mentioned this in another one of your pull requests. I think we should abstract all output to the console using some debug method. 

Also, since this is an error, isn't `console.error('foo');` a thing?

src/server/requestHandlers.js
```diff
@@ -59,43 +52,73 @@ function isValidPositiveIntegerString(x) {
 /******************************************************************************
 * Getters                                                                     *
 ******************************************************************************/
-exports.code = function code(request, response) {
+exports.codeByID = function codeByID(request, response) {
```
cdelahousse, line 55: I don't like this method name. If this is a getter, shouldn't it be 'getCodeById'. It's more explicit.

Also, I'm conflicted about about the capitalized D in `codeByID`. It's not camelCase and JS tends to lowercase the D (see [getElementById](https://developer.mozilla.org/en-US/docs/DOM/document.getElementById) ). On the other hand, I like ID more than Id.

src/server/requestHandlers.js
```diff
@@ -59,43 +52,73 @@ function isValidPositiveIntegerString(x) {
...
+    }
+    if (reply === null) {
+      return error(response, 404, 'Code not found');
+    }
+    return success(response, JSON.parse(reply));
+  });
 };
 
-exports.comment = function comment(request, response) {
+exports.commentsOnLine = function commentsOnLine(request, response) {
```
cdelahousse, line 72: Same thing here as above. If this is a getter, have 'get' as part of the method name.

src/server/requestHandlers.js
```diff
@@ -59,43 +52,73 @@ function isValidPositiveIntegerString(x) {
...
+    var out = [];
+    for (var i in reply) {
+      out.push(JSON.parse(reply[i]));
+    }
+    return success(response, out);
+  });
 };
 
-exports.comments = function comments(request, response) {
+exports.countComments = function countComments(request, response) {
```
cdelahousse: Your other comment related getter is commentsOnLine. That has comments as the first word. This, `countComments` has comments as the second word. Not only do I prefer `commentCount`, it's more consistent. It'll also mesh better with other method names as we expand our API (eg. messageCount, codeCount instead of countMessages, countCode). We'll have `<subject><thing-we-want>` instead of `<thing-we-want><subject>`.

src/server/requestHandlers.js
```diff
@@ -59,43 +52,73 @@ function isValidPositiveIntegerString(x) {
...
   var query = url.parse(request.url, true).query;
   var code_id = query.code_id;
-  Comment.findAll({where: {code_id: code_id}}).success(function(comments) {
-    return success(response, {code_id:code_id, comments:comments});
+  redis.smembers('comment:' + code_id + ':indices', function(err, reply) {
+    if (err !== null) {
+      return error(response, 500, 'Error while reading from database.');
```
cdelahousse, line 102: I think error() is a bit too generic a name. It could apply to a lot of other things. Maybe this should be errorResponse or something.

src/server/requestHandlers.js
```diff
@@ -109,14 +132,11 @@ exports.newcode = function newcode(request, response) {
...
   var id=uuid.v4();
-  Code.build({
-    uuid: id,
-    text: obj.text,
-    lang: obj.lang
-  }).save().success(function(code){
-    return success(response, code);
-  }).error(function(){
-    return error(response, 500, 'Error writing code to database');
+  redis.set('code:' + id, JSON.stringify(obj), function(err) {
```
cdelahousse: I think we should start using a replacer function argument with JSON.stringify. 

https://developer.mozilla.org/en-US/docs/Using_native_JSON#The_replacer_parameter

@asomayaji was speaking about whitelists and a replace function would be a super easy implementation.

src/server/requestHandlers.js
```diff
@@ -140,26 +160,16 @@ exports.newcomment = function newcomment(request, response) {
   if (Number(fields.line_start) > Number(fields.line_end)) {
```
cdelahousse, line 156: Number will detect many different types of numbers and then convert them to JS numbers (eg. Number("2e1") === 20). `parseInt(varName, 10)`  will parse base 10 numbers until it detects a non-digit (eg. parseInt("2e1", 10) === 2).

I think parseInt( ... , 10) is a better bet.

http://stackoverflow.com/questions/4090518/string-to-int-use-parseint-or-number

bin/stop_dev.sh
```diff
@@ -12,10 +12,11 @@
...
+DB_PID="var/db.pid"
 
 ######################################################################
-# Stop server
-
-kill -9 `cat $PID_FILE`
+# Stop server and db
 
+kill -9 `cat $SERVER_PID`
+kill -9 `cat $DB_PID`
```
psimonyi: How about `kill -SIGKILL` (for self-documentation)?
Does it matter that this gives the database no chance to clean up (e.g. write to a file)?

psimonyi: In Bash, `$(< file)` does basically the same thing as ``cat file`` (without actually executing `cat`). Not sure whether you'd prefer to use it or not. Consider this comment free trivia.

etc/redis.conf
```diff
@@ -0,0 +1,571 @@
...
+#
+# units are case insensitive so 1GB 1Gb 1gB are all the same.
+
+# By default Redis does not run as a daemon. Use 'yes' if you need it.
+# Note that Redis will write a pid file in /var/run/redis.pid when daemonized.
+daemonize no
+
+# When running daemonized, Redis writes a pid file in /var/run/redis.pid by
+# default. You can specify a custom pid file location here.
+pidfile var/redis.pid
```
psimonyi: The start script also writes the PID do a file with a different name (`db.pid`)...

etc/redis.conf
```diff
@@ -0,0 +1,571 @@
...
+# Specify the syslog identity.
+# syslog-ident redis
+
+# Specify the syslog facility. Must be USER or between LOCAL0-LOCAL7.
+# syslog-facility local0
+
+# Set the number of databases. The default database is DB 0, you can select
+# a different one on a per-connection basis using SELECT <dbid> where
+# dbid is a number between 0 and 'databases'-1
+databases 16
```
psimonyi, line 84: Why 16? Do we ever use more than one? Does this matter?

etc/redis.conf
```diff
@@ -0,0 +1,571 @@
...
+# This makes the format more resistant to corruption but there is a performance
+# hit to pay (around 10%) when saving and loading RDB files, so you can disable it
+# for maximum performances.
+#
+# RDB files created with checksum disabled have a checksum of zero that will
+# tell the loading code to skip the check.
+rdbchecksum yes
+
+# The filename where to dump the DB
+dbfilename dump.rdb
```
psimonyi: "dump.rdb"? Seems fine for dev mode, but this is the same configuration for real use, so maybe there's a name that makes it sound less... disposable.

src/server/requestHandlers.js
psimonyi: I support the proposal of the honourable member from Sandy Hill. (It's not because I particularly care about the function name, but because I think /do/commentCount looks better than /do/countComments -- it's more declarative than imperative, as a GET URL should be, and as @cdelahousse points out, it'll probably be more obviously better when we have "count" operations on other kinds of things.)

src/server/requestHandlers.js
psimonyi, line 102: Given the context (request handlers), it seems fairly reasonable (and pleasantly short!) to me. Anyway, if you want to change it, be sure to change `success` accordingly.

src/server/requestHandlers.js
psimonyi, line 55: @cdelahousse This function is a request handler (not exactly a "getter" in the usual sense). It's named after the URL pattern it handles. I'm not really a fan of `/do/codeByID`, but I do like having the function name match the URL. But if we're going to discuss the name of the function, I think we should optimize for nice URLs.

src/server/requestHandlers.js
```diff
@@ -59,43 +52,73 @@ function isValidPositiveIntegerString(x) {
...
+  }
+  redis.lrange('comment:' + id + ':' + line, 0, -1, function(err, reply) {
+    if (err !== null) {
+      return error(response, 500, 'Error while reading from database.');
+    }
+    if (reply === null) {
+      return error(response, 404, 'Comment not found');
+    }
+    var out = [];
+    for (var i in reply) {
```
psimonyi: If reply is an array, this would be better as a forEach call. (for/in doesn't guarantee any ordering)

src/server/requestHandlers.js
```diff
@@ -59,43 +52,73 @@ function isValidPositiveIntegerString(x) {
...
+  redis.smembers('comment:' + code_id + ':indices', function(err, reply) {
+    if (err !== null) {
+      return error(response, 500, 'Error while reading from database.');
+    }
+    if (reply === null) {
+      return error(response, 404, 'Comments not found.');
+    }
+    var multi = redis.multi();
+    for (var i in reply) {
+      if (reply.hasOwnProperty(i)) {
```
psimonyi: Again, looping over array \rarrow forEach.

src/server/requestHandlers.js
```diff
@@ -59,43 +52,73 @@ function isValidPositiveIntegerString(x) {
...
+        console.log(i, reply[i]);
+        multi.llen('comment:' + code_id + ':' + reply[i]);
+      }
+    }
+    multi.exec(function(err, replies) {
+      var out = {};
+      replies.forEach(function(value, index) {
+        console.log(index, reply[index], value);
+        out[reply[index]] = value;
+      });
```
psimonyi, line 115: The relationship between `reply` and `replies` is not very clear. I'm not sure what to suggest.

src/server/requestHandlers.js
psimonyi, line 156: It checks for isValidPositiveIntegerString a few lines before this, so I don't think it matters whether we use Number or parseInt as long as the validator works right. However, I think it would be ideal if we could combine the validation and conversion: so you get a tainted string from the network, and say `var myNumber = toPositiveInteger(string, onFailure)` or something. We should look into this for the security audit #32.

etc/redis.conf
```diff
@@ -0,0 +1,571 @@
...
+# Note that Redis will write a pid file in /var/run/redis.pid when daemonized.
+daemonize no
+
+# When running daemonized, Redis writes a pid file in /var/run/redis.pid by
+# default. You can specify a custom pid file location here.
+pidfile var/redis.pid
+
+# Accept connections on the specified port, default is 6379.
+# If port 0 is specified Redis will not listen on a TCP socket.
+port 6379
```
psimonyi, line 25: For security and ease of deployment, unix fifos are probably preferable to sockets (since you don't have to worry about firewalls). But it's not clear how to do this from redis's Node bindings. This question can be pushed off to #32.

src/server/requestHandlers.js
psimonyi: Apparently in Node, `console.error` just [prints to stderr](http://nodejs.org/api/stdio.html#stdio_console_error_data). (Since we direct stdout and stderr to the same file, nobody would see a difference.) But since it exists, it might be good to use it, so we can take advantage of any changes, and since it's more self-documenting. Or maybe in the future we'll use some logging module that makes the difference more significant.

src/server/requestHandlers.js
psimonyi: todo: #32

psimonyi:
I think that's all for now.

src/server/requestHandlers.js
cdelahousse, line 55: You're totally right.

If my understanding of RESTful APIs is correct, we both write and read from the same resource depending on the HTTP verbs.  `newcomment` and `commentOnLine` shouldn't be served at two different resources (`/do/newcomment`, `/do/commentOnLine`). They should both be served at `/do/comment` or `/do/comment/line/` and the HTTP verb and other headers should dictate what we want to do. We shouldn't have separate getter and setter resources.

src/client/scripts/code.js
spratt, line 16: Without further comment, I'm considering this thread

[resolved]

INSTALL
spratt, line 14: This isn't really an issue for this pull request, I'm considering this thread

[resolved]

src/client/scripts/comment.js
spratt: These aren't methods, but you are correct because of subsection 14 of the Javascript section of the coding standard.

src/client/scripts/main.js
spratt, line 69: None of the lines I edited had double quotes.  We will do a cleanup later on all the existing code that breaks the new coding standard.

spratt, line 69: I'm considering this

[resolved]

src/client/scripts/main.js
spratt: Agreed, this breaks js.15 of the coding standard.

src/server/requestHandlers.js
spratt, line 55: I think this is fine for now.  It works and it meets the coding standard.  If you'd like to clean it up, I would be super happy about that.

src/server/requestHandlers.js
spratt, line 72: I think this is fine for now.

src/server/requestHandlers.js
spratt: Fair.

src/server/requestHandlers.js
spratt, line 102: No.

src/server/requestHandlers.js
spratt: [resolved]

src/server/requestHandlers.js
spratt, line 156: This was not part of the diff.  Make a separate issue.

[resolved]

bin/stop_dev.sh
spratt: Fair.

etc/redis.conf
spratt: This line is ignored because we're not running in daemonized mode.

etc/redis.conf
spratt, line 84: Default.

etc/redis.conf
spratt: default

src/server/requestHandlers.js
spratt: Fair.

src/server/requestHandlers.js
spratt: Fair.

src/server/requestHandlers.js
spratt, line 115: I'll clarify in code.

etc/redis.conf
spratt, line 25: I agree.  On our shared server, we also have no guarantee the port will be available.

spratt, line 25: [resolved]

src/client/scripts/main.js
spratt: This is actually resolved.  Github is being dumb.

[resolved]

etc/redis.conf
spratt, line 84: [resolved]

etc/redis.conf
spratt: [resolved]

src/server/requestHandlers.js
spratt, line 55: This isn't resolved, exactly, but Peter considers this

[nonblocking]

src/server/requestHandlers.js
spratt, line 72: This isn't resolved, exactly, but Peter considers this

[nonblocking]

psimonyi:
Looks good. There are a couple of non-blocking things that might deserve more consideration in separate issues.

src/server/requestHandlers.js
psimonyi, line 156: #32

PR #111, opened 2013-04-01
Persona back-end integration

### LINKED ISSUE #39: Persona integration
_Imported issue by **[spratt](http://github.com/spratt)** from 2012-07-03T13:02:22Z_
_Originally opened as https://github.com/SteamedPears/CodeReviewClientMaster/issues/14_

---

Add Open ID integration for users who wish to log in.  I would like anonymous submissions to remain an option, so we can compete with paste bin.

P.S. Persona may be better than Open ID.

### LINKED ISSUE #104: Our page layout sucks
I hate how the tutorial button runs right up against the code review text box.  I think we should put the tutorial button in the top right, aligned vertically with the logo on the left.

spratt:
We now perform all the necessary steps on the server-side to fully authenticate users with the Persona protocol.

Once a user is authenticated, their email is stored in the session variables.  Users now comment either as "Anonymous" (if not signed in) or as their email (if signed in).

closes #39
closes #104


spratt:
Every weekend I ask you guys to pull another fuck-you-sized request.  I am sorry about this.

src/client/index.html
```diff
@@ -4,17 +4,31 @@
...
     <link href="styles/view.css" rel="stylesheet">
     <link href="scripts/lib/intro.js-0.2.1/minified/introjs.min.css" rel="stylesheet">
     <link href="scripts/lib/bootstrap/css/bootstrap.min.css" rel="stylesheet" media="screen">
     <script data-main="scripts/main" src="scripts/lib/require.js"></script>
   </head>
   <body>
-    <h1><a href="index.html">CodeReview</a></h1>
+    <h1 id="title"><a href="index.html">CodeReview</a></h1>
+      Try our
+      <a href="javascript:(function(){window.codeReview&&window.codeReview.exit();var a=document.createElement('script');a.src='//review.steamedpears.com/bookmarklet/bookmarklet.js';(document.head||document.body).appendChild(a)})();">bookmarklet</a>
```
cdelahousse, line 16: Good call, moving the bookmarklet up.

src/client/scripts/login.js
```diff
@@ -0,0 +1,79 @@
...
+  var label_span = null;
+  var user_display = null;
+  var button = null;
+  $(document).ready(function() {
+    user_display = $('#comment-user, #user-email');
+    button = $('#persona-button-text');
+    label_span = $('#user-label');
+  });
+
+  $.getScript('https://login.persona.org/include.js', ready);
```
cdelahousse: Couldn't we just pull this in using RequireJS instead of with jQuery? 
http://requirejs.org/docs/api.html#config-paths

src/client/scripts/login.js
```diff
@@ -0,0 +1,79 @@
...
+  $('#persona-login').click(function login_persona_click() {
+    navigator.id.request();
+    return false;
+  });
+  $('#persona-logout').click(function login_persona_click() {
+    navigator.id.logout();
+  });
+  function ready() {
```
cdelahousse: Is this a good idea to separate the `ready` function from where it's set as a callback on line 21. Why not just define this inline?

src/client/scripts/view.js
```diff
@@ -123,6 +123,7 @@ define([
     $('#line-end-num').text(end+1);
     editor.setDiffSelected(start,end);
     var pos = editor.getLinePosition(start);
+    console.log(pos);
```
cdelahousse: Is this debug code? You may want to strip this out.

src/server/dev_mode.js
```diff
@@ -12,15 +12,14 @@ var proxy = require('proxy-middleware');
...
+module.exports = function(host, serverPort, clientPort) {
   // Configuration
   var staticPort = 8888;
   var staticDirectory = '../client/';
 
-  var proxyPort  = 8080;
   var proxyRoutes = {
-    '/do' : 'http://localhost:' + serverPort,
-    '/'   : 'http://localhost:' + staticPort
+    '/do' : 'http://' + host + ':' + serverPort,
```
cdelahousse, line 21: Nice.

src/server/index.js
```diff
@@ -59,7 +65,9 @@ for (var route in getRoutes) {
   app.use(route, {handle:getRoutes[route]});
 }
 
-app.use(connect.bodyParser());
+app.use(connect.bodyParser())
+  .use(connect.cookieParser())
+  .use(connect.session({secret: 'keyboard kitties'}));
```
cdelahousse, line 70: Giggles.

src/server/requestHandlers.js
```diff
@@ -121,14 +122,19 @@ exports.commentCount = function commentCount(request, response) {
 /******************************************************************************
 * Setters                                                                     *
 ******************************************************************************/
-exports.newcode = function newcode(request, response) {
+function newcode(request, response) {
```
cdelahousse: I'm not a fan of removing the exports assignments form the function declaration. If you keep it, I know that as I read the code, I can expect this to be a public facing API instead of learning about this near the bottom of the file.

spratt: The reason I did this was to pass in the host and clientPort arguments, and I didn't want to put all of the functions inside of the function I used for that.  See line 199.

cdelahousse: Ah.

spratt: I'm not super happy with it either.  Suggestions?

src/client/scripts/login.js
spratt: It was part of @psimonyi's original implementation.  I think I'd prefer to use RequireJS for consistency, so I'll see if I can get it working that way instead.

src/client/scripts/login.js
spratt: Part of the original implementation.  This was probably more similar to our conventions at the time.

src/client/scripts/view.js
spratt: Good find.

src/server/index.js
spratt, line 70: There's probably a good convention for what we should use as our secret key.  This is basically connect.js's default.

src/server/requestHandlers.js
spratt: I have a better idea.  Commit incoming.

src/client/index.html
spratt, line 16: Thanks!

spratt:
[CRITICAL] I've found a major bug that causes the db to lose the lang attribute on code objects.  Do not merge this until fixed.

spratt:
Fixed.  The problem was a cascade of bugs.

First, if you were visiting a piece of code you hadn't just submitted, the language select box would be unpopulated.  If you expanded comments on a line, it would submit the code form again.  Our `/do/newcode` request wasn't checking for a valid lang, so it would succeed and create a new code object.  This would change the url of the page to the new ID.  Now, if you refreshed the page you would get the new code with unspecified lang, which made it look like the db was dropping the lang attribute.

Fixes:
- `/do/newcode` now checks if the lang attribute is properly set.
- Expanding the comments on a line doesn't submit the code form.

src/server/dev_mode.js
spratt, line 21: I like this better...but it will only ever be called with host='localhost', since this is just for dev mode.  Oh well.

src/server/requestHandlers.js
```diff
@@ -156,17 +170,89 @@ exports.newcomment = function newcomment(request, response) {
...
+* In the event of implementing more than just persona for authentication, you *
+* should replace most of the following code with a library like               *
+*                                                                             *
+* passport           http://passportjs.org/                                   *
+* passport-browserid https://github.com/jaredhanson/passport-browserid        *
+*                                                                             *
+******************************************************************************/
+
+module.exports = function(host, clientPort) {
+  public_api.login = function login(request, response) {
```
spratt, line 213: This is the most complex part of the back-end now.  I'm not sure how I feel about this.

cdelahousse:
It looks good and you've fixed the bug. Can I merge this bitch now?

spratt:
May as well let @psimonyi have a look.

src/client/index.html
```diff
@@ -4,17 +4,31 @@
...
+        <span id="user-email"></span>
+      </span>
+      <button id="persona-login" class="persona-button">
+        <span id="persona-button-text">Sign in</span>
+      </button>
+      <button id="tutorial-btn" class="btn btn-success">
+        Tutorial
+      </button>
+    </aside>
+    <br clear="all">
```
psimonyi, line 30: :(

src/server/requestHandlers.js
```diff
@@ -1,9 +1,15 @@
 // Libraries
+var https = require('https');
 var url = require('url');
 var uuid = require('node-uuid');
 var db = require('redis').createClient();
 
 /******************************************************************************
+* The methods on this object will be exported as the public api.              *
+******************************************************************************/
+var public_api = {};
```
psimonyi, line 10: Does one line of code really need a 3-line comment heading? I thought the starry stuff was for sections of code, but this comment isn't about a section.

cdelahousse, line 10: Agreed

src/server/requestHandlers.js
```diff
@@ -156,17 +173,88 @@ exports.newcomment = function newcomment(request, response) {
...
+module.exports = function(host, clientPort) {
+  public_api.login = function login(request, response) {
+    if(request === null ||
+       request.body === null ||
+       request.body.assertion === null) {
+      return error(response, 400, 'Invalid assertion');
+    }
+    var assertion = request.body.assertion;
+    var content = 'assertion=' + assertion + '&audience=http://';
+    content += host + ':' + clientPort;
```
psimonyi: In Python, I'd have `'assertion={asrt}&audience=http://{host}:{port}'.format(asrt=assertion, host=host, port=port)`. Isn't there something like this in JS?

src/server/requestHandlers.js
```diff
@@ -156,17 +173,88 @@ exports.newcomment = function newcomment(request, response) {
...
+
+module.exports = function(host, clientPort) {
+  public_api.login = function login(request, response) {
+    if(request === null ||
+       request.body === null ||
+       request.body.assertion === null) {
+      return error(response, 400, 'Invalid assertion');
+    }
+    var assertion = request.body.assertion;
+    var content = 'assertion=' + assertion + '&audience=http://';
```
psimonyi: What if someone provides an assertion like `1&audience=http://mysite.com`?

src/server/requestHandlers.js
```diff
@@ -156,17 +173,88 @@ exports.newcomment = function newcomment(request, response) {
...
+    var auth_request = https.request({
+      host: 'verifier.login.persona.org',
+      port: 443,
+      path: '/verify',
+      method: 'POST',
+      headers: {'Content-Length': content.length,
+                'Content-Type': 'application/x-www-form-urlencoded'}
+    },function(auth_response) {
+      var data_str = '';
+      // assume uft8
```
psimonyi, line 231: Can't we use Connect's body_parser thing here too? I thought it handled the encoding and waiting for all data things.

http_api.md
```diff
@@ -0,0 +1,187 @@
...
+# Setters
+
+## Code
+
+URL: `/do/newcode`
+
+Required parameters:
+
+* `text`: The (code) text of the code object.
+* `lang': The programming language in which the text is written.
```
psimonyi: Closing quote: Markdown, not LaTeX.

src/client/index.html
spratt, line 30: What's wrong with this?

src/client/index.html
psimonyi, line 16: This URL exists elsewhere, right? We should somehow pull it from wherever it lives, or else remove it from the old place and make this canonical. IOW, this thing should exist in one place in the repository. [non-blocking, file-a-bug]

src/client/index.html
psimonyi, line 30: [non-blocking]

src/client/scripts/login.js
```diff
@@ -0,0 +1,81 @@
...
+  
+  $(document).ready(function() {
+    user_display = $('#comment-user, #user-email');
+    button = $('#persona-button-text');
+    label_span = $('#user-label');
+  });
+
+  require([
+    'https://login.persona.org/include.js'
+  ], function persona_ready() {
```
psimonyi: Why not just include this in the dependencies for `define` and save on a callback and indentation level?

src/client/scripts/login.js
```diff
@@ -0,0 +1,81 @@
...
+    }
+  }
+  function logout_success(data) {
+    if(label_span !== null) {
+      label_span.hide();
+    }
+    if(user_display !== null) {
+      user_display.text('Anonymous');
+    }
+    if(button !== null) {
```
psimonyi: Could we have spaces after the `if`s please?

src/server/requestHandlers.js
spratt, line 231: body_parser is meant to go into the connect call chain and parse the body of an incoming request, not the body of an incoming response to a request we sent.  It might be able to do it, but it's not what it's intended for.

src/server/requestHandlers.js
```diff
@@ -121,14 +129,22 @@ exports.commentCount = function commentCount(request, response) {
 /******************************************************************************
 * Setters                                                                     *
 ******************************************************************************/
-exports.newcode = function newcode(request, response) {
+public_api.newcode = function newcode(request, response) {
   // do some basic validation
-  var obj = request.body;
-  if (obj === null || !isValidString(obj.text)) {
+  var fields = request.body;
+  if (fields === null || !isValidString(fields.text)) {
```
psimonyi: Not just `if (!fields || ...`? As it is, the empty string is valid code. And what if request.body is undefined?

src/server/requestHandlers.js
```diff
@@ -94,9 +99,12 @@ exports.commentsOnLine = function commentsOnLine(request, response) {
   });
 };
 
-exports.commentCount = function commentCount(request, response) {
+public_api.commentCount = function commentCount(request, response) {
   var query = url.parse(request.url, true).query;
   var code_id = query.code_id;
+  if(code_id === undefined) {
+    return error(response, 400, 'Invalid code id');
```
psimonyi, line 118: Shouldn't this return invalid code id when the code id isn't in the database too? It seems there should be a difference between a code id that doesn't exist, and a code id that exists but has no comments.

src/server/requestHandlers.js
psimonyi, line 213: Perhaps this belongs in its own module. [non-blocking]

src/server/requestHandlers.js
psimonyi: We should probably URL-encode the assertion.

src/client/index.html
spratt, line 30: Most of the HTML is about presentation?  It's basically a structure into which we inject our content.

src/server/requestHandlers.js
```diff
@@ -156,17 +171,88 @@ exports.newcomment = function newcomment(request, response) {
...
+      var data_str = '';
+      // assume uft8
+      auth_response.setEncoding('utf8');
+      auth_response.on('data',function(data) {
+        data_str += data;
+      });
+      auth_response.on('end',function() {
+        var data_ob = JSON.parse(data_str);
+        if (auth_response.statusCode === 200 &&
+            data_ob !== null &&
```
psimonyi: Apparently `JSON.parse` will throw an exception for invalid JSON. It would only return `null` if the JSON text is acutally `null`. [Reference](https://developer.mozilla.org/en-US/docs/JavaScript/Reference/Global_Objects/JSON/parse)

src/server/requestHandlers.js
psimonyi, line 213: Made a bug.

src/server/requestHandlers.js
psimonyi, line 118: This got changed somewhat in another commit, and this behaviour is consistent within CR, so it's fine.

src/server/requestHandlers.js
```diff
@@ -156,17 +181,94 @@ exports.newcomment = function newcomment(request, response) {
...
+*                                                                             *
+* passport           http://passportjs.org/                                   *
+* passport-browserid https://github.com/jaredhanson/passport-browserid        *
+*                                                                             *
+******************************************************************************/
+
+module.exports = function(host, clientPort) {
+  public_api.login = function login(request, response) {
+    if(request === null ||
+       request.body === null ||
```
psimonyi: With Connect, it appears request.body will never be null. If it's any falsy value, body_parser sets it to an empty object. I think `if (!request.body.assertion)` will do.

src/client/index.html
```diff
@@ -4,17 +4,31 @@
...
+      Try our
+      <a href="javascript:(function(){window.codeReview&&window.codeReview.exit();var a=document.createElement('script');a.src='//review.steamedpears.com/bookmarklet/bookmarklet.js';(document.head||document.body).appendChild(a)})();">bookmarklet</a>
+      by dragging the link to your bookmarks bar.
+    <aside id="user-id">
+      <span id="user-label">
+        Signed in as:
+        <span id="user-email"></span>
+      </span>
+      <button id="persona-login" class="persona-button">
+        <span id="persona-button-text">Sign in</span>
```
psimonyi, line 24: Why do you need this extra span?

psimonyi, line 24: So it's my code. We still don't need it.

spratt, line 24: CSS.

psimonyi, line 24: Oops. Right.

src/server/requestHandlers.js
```diff
@@ -130,37 +156,110 @@ exports.newcode = function newcode(request, response) {
...
+        data_str += data;
+      });
+      auth_response.on('end',function() {
+        try {
+          var data_ob = JSON.parse(data_str);
+        } catch(err) {
+          return error(response, 500, 'Invalid verification server response');
+        }
+        if (auth_response.statusCode === 200 &&
+            data_ob && data_ob.status &&
```
psimonyi: There's no need to check that both `data_ob.status` is not falsy and that it is `'okay'`.

psimonyi:
`/do/login` returns a 400 Bad Request. The log file shows `===ERROR=== 400 Invalid assertion`, and also a whole bunch of errors connecting to redis.

psimonyi:
The database server doesn't seem to be started by the scripts in this branch. Is this affecting `/do/login`?

spratt:
bin/start_dev.sh seems to start the database for me.  But I am also getting a 400 on /do/login.  Investigating.

spratt:
I have inverted logic in the sanity check for /do/login.  Fixing now.

Issue #45, opened 2013-02-05
Production Deployment


spratt:
Need to figure out how to easily deploy new versions of the app to the production server.  Ideally this should be as automated as possible.

bheesham:
I think that the scripts you prepared would do fine. Something along those lines.

spratt:
I like using nginx to serve the static files, and webfaction requires the static files to be in a different directory.  Also, production will be using mysql or postgres.

cdelahousse:
What'll be the process to deploy? Can anybody to it? When is does the build get to a deployable state?

spratt:
@bheesham It would be really awesome if you could finish this when you have time.  

Check out this article on deploying node.js to webfaction to see the issues we need to solve: http://shkfon.tumblr.com/post/27178918675/real-world-nodejs-part-1

psimonyi:
Is this resolved by #126? @spratt

spratt:
Yup.

PR #112, opened 2013-04-01
Server install script

bheesham:
Works without having to modify the PATH variable.


bin/install_server.sh
```diff
@@ -0,0 +1,40 @@
...
+######################################################################
+
+ROOT_DIR=`pwd`
+TIME=`date +%Y-%m-%dT%H%M%S%Z`
+
+######################################################################
+# Configuration
+BIN_DIR=$ROOT_DIR/bin/exe
+
+HOME_FOLDER=/home/bheesham/
```
spratt: This is less than ideal.  It would be best if it could work on any server without modification. [non-blocking]

.gitignore
```diff
@@ -1,3 +1,18 @@
-npm-debug.log
-*.DS_Store
```
spratt, line 2: This line is important for my development environment.  [blocking]

spratt:
This is a nice install script.  Note that it doesn't address all the requirements for #45 such as:
- updated INSTALL documentation
- copying the client-side to a target directory for static serving

.gitignore
```diff
@@ -1,3 +1,18 @@
-npm-debug.log
```
spratt, line 1: This line is important to ignore because npm generates this file if it fails with an error, for example a build error.  [blocking]

INSTALL.md
```diff
@@ -4,6 +4,12 @@ Installation Instructions
 Development
 -----------
 
+The setup script (```bin/setup_dev.sh```) will download node and npm,
+from there node can install sqlite. To use the script, edit it and
```
spratt: We're not using sqlite any more.  [blocking]

INSTALL.md
```diff
@@ -4,6 +4,12 @@ Installation Instructions
 Development
 -----------
 
+The setup script (```bin/setup_dev.sh```) will download node and npm,
```
spratt: This script doesn't exist.  [blocking]

bin/install_server.sh
spratt: Isn't there a bash variable for the user's home folder already?  Like ~?

spratt:
You seem to have included your sqlite database in the commit:  src/server/models/data.db which should be removed before merging.  [blocking]

bin/install_server.sh
bheesham: Yeah. Run a bash script as root, and that changes to /root, with all of the files located elsewhere.
I avoided that for a reason.

There could possibly be a better way to do this.

bin/install_server.sh
```diff
@@ -0,0 +1,40 @@
...
+#
+# This script should install nodejs, and use the default configuration
+# for npm.
+######################################################################
+
+ROOT_DIR=`pwd`
+TIME=`date +%Y-%m-%dT%H%M%S%Z`
+
+######################################################################
+# Configuration
```
cdelahousse, line 15: See the comment header proposal in the Coding Standard. [no-biggy]

LICENSE
```diff
@@ -5,7 +5,7 @@ any purpose with or without fee is hereby granted, provided that the
 above copyright notice and this permission notice appear in all
 copies.
                                                                  
-THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL
+THE SOFTWARE IS PROVIDED ``AS IS'' AND THE AUTHOR DISCLAIMS ALL
```
cdelahousse, line 8: Are double quotes in text files bad?

INSTALL.md
```diff
@@ -4,6 +4,12 @@ Installation Instructions
...
+from there node can install sqlite. To use the script, edit it and 
+replace the two instances of ``CHANGEME'' with your preferred values.  
+
+**You will need g++ to be installed**.
```
cdelahousse: Can you not add this as a list item on the list nearby?

INSTALL.md
```diff
@@ -20,7 +26,8 @@ Instructions:
  1. open a terminal
  2. cd to the root directory of the project (containing this text
     file).
- 3. Run bin/start_dev.sh
+ 3. Run ```bin/setup_dev.sh```
+ 3. Run ```bin/start_dev.sh```
```
cdelahousse: I know markdown will convert this 3 to a 4 for numbered lists, but some of us will be viewing this in plain text. Please make this a proper list. [blocking]

bin/install_server.sh
```diff
@@ -0,0 +1,40 @@
...
+#
+# Project: Code Review
+# By:      Steamed Pears
+#
+# This script should install nodejs, and use the default configuration
+# for npm.
+######################################################################
+
+ROOT_DIR=`pwd`
+TIME=`date +%Y-%m-%dT%H%M%S%Z`
```
cdelahousse: Hmmm... Maybe we should standardize our time format in the coding standards?

INSTALL.md
spratt: Oh man...do NOT look at the source for the new version of the coding standard.  Every list item is number 1.

cdelahousse: But those live on github. You'll rarely be looking at the wiki source code. This INSTALL file WILL be read in plain text.

bin/install_server.sh
spratt: We really should.

spratt: Added to the proposals.  Also, we could make it a helper function and source it in.

psimonyi: +1

bin/install_server.sh
```diff
@@ -0,0 +1,40 @@
...
+echo Downloading node.js
+wget -q http://nodejs.org/dist/v$NODE_VERSION/$NODE_F.tar.gz
+
+echo Download complete, extracting node.js
+tar -xf $NODE_F.tar.gz -C $BIN_DIR/
+rm $NODE_F.tar.gz
+mv $NODE_F node
+
+echo Configuring npm
+$NODE_BINS/npm config list >> $HOME_FOLDER/.npmrc
```
psimonyi: Is this even needed? I don't think this actually does anything. (And then we don't need HOME_FOLDER).

LICENSE
bheesham, line 8: No not really. That's just how you represent the left and right quotes in plain text.
Tex converts those to "real" quotes, not that we're using tex.

bin/install_server.sh
bheesham: I thought we needed it. As it turns out, we don't.
NPM will just use the default stuff. I guess it changed since I first used it?

.gitignore
```diff
@@ -1,5 +1,24 @@
...
+lib-cov
+*.seed
+*.log
+*.csv
+*.dat
+*.out
+*.pid
+*.gz
+*.swp
+*.swo
```
psimonyi: Do we need all this again?

PR #113, opened 2013-04-01
Production build script

cdelahousse:
- uses r.js to optimize the client
- uses uglify to minify the bookmarklet.


.gitignore
```diff
@@ -1,3 +1,4 @@
 npm-debug.log
 *.DS_Store
 *~
+production*
```
spratt: This is weird.  Aren't we just ignoring production/ and production_tmp/ ?  I think we should ignore those explicitly.  [non-blocking]

bin/build_production_all.sh
```diff
@@ -0,0 +1,84 @@
...
+# Configuration                                                               #
+###############################################################################
+
+CLIENT_DIR="$PWD/src/client"
+CLIENT_SCRIPTS_DIR="$CLIENT_DIR/scripts"
+
+BOOKMARKLET_DIR="$PWD/src/bookmarklet"
+
+#Allow production-build directory to be specified by command line args
+TARGET="$PWD/$1"
```
spratt, line 16: This still feels like it should be a Makefile.  [non-blocking]

src/client/scripts/app.build.js
```diff
@@ -1,20 +1,8 @@
...
-    modules: [
-        //Optimize the application files. jQuery is not 
-        //included since it is already in require-jquery.js
-        {
-            name: "main"
-        }
-    ]
+  appDir: '../', //Where the client is located
+  baseUrl: 'scripts/', //Where look for scripts by default
+  dir: '../../../production', //Where to built the optimized project
```
spratt: Where to build\* [non-blocking]

src/client/scripts/app.build.js
```diff
@@ -1,20 +1,8 @@
...
-
-    modules: [
-        //Optimize the application files. jQuery is not 
-        //included since it is already in require-jquery.js
-        {
-            name: "main"
-        }
-    ]
+  appDir: '../', //Where the client is located
+  baseUrl: 'scripts/', //Where look for scripts by default
```
spratt: Where to\* look [non-blocking]

src/client/scripts/app.build.js
```diff
@@ -1,20 +1,8 @@
...
-    },
-
-    modules: [
-        //Optimize the application files. jQuery is not 
-        //included since it is already in require-jquery.js
-        {
-            name: "main"
-        }
-    ]
+  appDir: '../', //Where the client is located
```
spratt: What is the client in this context? [blocking]

src/bookmarklet/README.md
```diff
@@ -14,12 +14,15 @@ move everything to them to where they can be served.
 Run `make build` to build the test-build version of the bookmarklet and vendors
 and move them to where files are are served. 
 
+Run `make build-local` to build the project locally.
```
spratt: This makes it seem like the normal build is remote.  I really don't know what you mean here.  [blocking]

bin/build_production_all.sh
```diff
@@ -0,0 +1,84 @@
...
+if [ -z "$1" ]
+then
```
spratt: the convention is to put `then` on the same line as `if`.  Such as in the following example:

```
if [ condition ]; then
```

Not sure if this should be [blocking].  It's a minor issue, but it breaks convention.

bin/build_production_all.sh
cdelahousse, line 16: I would rather not

spratt, line 16: Okay, I can rewrite it later.  Any shell file whose entire function is to generate files is better written as a Makefile.

bin/build_production_all.sh
```diff
@@ -0,0 +1,84 @@
...
+# Sanity check                                                                #
+###############################################################################
+
+for DIR in "$CLIENT_DIR" "$BOOKMARKLET_DIR"; do
+  test -d "$DIR"
+  exitIfFailed "Cannot find" `basename "$DIR"` ". Try running this \
+    script from the project root."
+done
+
+type uglifyjs >/dev/null 2>&1
```
spratt, line 41: You can do this simpler with `type uglifyjs &> /dev/null`

Issue #117, opened 2013-04-02
Use less locking in DB


spratt:
I even know how to do this - http://redis.io/commands/hincrby solves the worst of the situations I use locking for.

PR #118, opened 2013-04-02
Now using less locking in the db

spratt:
We no longer use locking for getting comment counts, only for creation of new comments.

### DIFF
--- a/src/server/requestHandlers.js
+++ b/src/server/requestHandlers.js
@@ -147,17 +147,15 @@ exports.newcomment = function newcomment(request, response) {
     return error(response, 400, 'Invalid line numbers');
   }
   var comment_key = 'comment:' + fields.code_id + ':' + fields.line_start;
-  return db.lpush(comment_key, JSON.stringify(fields), function(err) {
+  var transaction = db.multi();
+  transaction.lpush(comment_key, JSON.stringify(fields));
+  var indices_key = 'comment:' + fields.code_id + ':indices';
+  transaction.hincrby(indices_key, fields.line_start, 1);
+  return transaction.exec(function(err) {
     if (err !== null) {
       return error(response, 500, 'Error writing comment to database.');
     }
-    var indices_key = 'comment:' + fields.code_id + ':indices';
-    return db.hincrby(indices_key, fields.line_start, 1, function(err) {
-      if (err !== null) {
-        return error(response, 500, 'Error incrementing count in database.');
-      }
-      return success(response, fields);
-    });
+    return success(response, fields);
   });
 };
 
--- a/src/server/requestHandlers.js
+++ b/src/server/requestHandlers.js
@@ -147,7 +147,7 @@ exports.newcomment = function newcomment(request, response) {
     return error(response, 400, 'Invalid line numbers');
   }
   var comment_key = 'comment:' + fields.code_id + ':' + fields.line_start;
-  return db.lpush(key, JSON.stringify(fields), function(err) {
+  return db.lpush(comment_key, JSON.stringify(fields), function(err) {
     if (err !== null) {
       return error(response, 500, 'Error writing comment to database.');
     }
--- a/src/server/requestHandlers.js
+++ b/src/server/requestHandlers.js
@@ -97,24 +97,14 @@ exports.commentsOnLine = function commentsOnLine(request, response) {
 exports.commentCount = function commentCount(request, response) {
   var query = url.parse(request.url, true).query;
   var code_id = query.code_id;
-  db.smembers('comment:' + code_id + ':indices', function(err, indices) {
+  db.hgetall('comment:' + code_id + ':indices', function(err, indices) {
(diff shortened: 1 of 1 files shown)


src/server/requestHandlers.js
```diff
@@ -156,17 +146,21 @@ exports.newcomment = function newcomment(request, response) {
...
-      return success(response, fields);
-    });
+  function comment_done(err) {
+    if (err !== null) {
+      return error(response, 500, 'Error writing comment to database.');
+    }
+    db.hincrby('comment:' + fields.code_id + ':indices', fields.line_start, 1,
+               hincr_done);
+  }
+  function hincr_done(err) {
```
cdelahousse: I'm assuming you didn't declare `comment_done` inline as an anon function on line 163 because you did not want to have 3 levels of call back declarations (with `hincr_done`) and indentation. I was wondering with you could at least declare `hincr_done` within `comment_done`. I think it would express the hierarchy of callbacks a bit more effectively. The way you have it now, it's hard to know what calls what.

cdelahousse: Totally [non-blocking] by the way.

spratt: The indentation gets horrible when hincr_done is inline.  I feel like they're close enough for it to be clear.

cdelahousse: The reason I made the comment is because I paused while reading your code

spratt: Alright, fixed.

src/server/requestHandlers.js
```diff
@@ -156,17 +146,19 @@ exports.newcomment = function newcomment(request, response) {
   if (Number(fields.line_start) > Number(fields.line_end)) {
     return error(response, 400, 'Invalid line numbers');
   }
-  // upon successfully saving comment, this function will update comment indices
-  db.multi()
-    .lpush('comment:' + fields.code_id + ':' + fields.line_start,
-           JSON.stringify(fields))
-    .sadd('comment:' + fields.code_id + ':indices', fields.line_start)
-    .exec(function(err) {
+  var comment_key = 'comment:' + fields.code_id + ':' + fields.line_start;
```
cdelahousse, line 153: This variable doesn't seem to be used.

src/server/requestHandlers.js
```diff
@@ -156,17 +146,19 @@ exports.newcomment = function newcomment(request, response) {
...
+  return db.lpush(comment_key, JSON.stringify(fields), function(err) {
+    if (err !== null) {
+      return error(response, 500, 'Error writing comment to database.');
+    }
+    var indices_key = 'comment:' + fields.code_id + ':indices';
+    return db.hincrby(indices_key, fields.line_start, 1, function(err) {
```
cdelahousse: If db.hincrby fails, the indices will not be incremented, but the new code will still be inputted into redis? Am I right? Should both inputting code AND incrementing indices fail if `db.hincrby` fails?

I don't quite understand how redis works, so I may be off base. I'm assuming that `db.hincrby` and `db.lpush` are separate atomic actions.

src/server/requestHandlers.js
```diff
@@ -130,7 +120,7 @@ exports.newcode = function newcode(request, response) {
   var id=uuid.v4();
   db.set('code:' + id, JSON.stringify(obj), function(err) {
     if (err !== null) {
-      return error(response, 500, 'Error while writing to database.');
+      return error(response, 500, 'Error while writing code to database.');
```
cdelahousse, line 125: I like that we're becoming more specific with our error messages.

src/server/requestHandlers.js
spratt: They are individually atomic actions.  If hincrby fails, we would be inconsistent by one comment count.

We could make them part of a multi transaction as they were before and then both actions would be a single atomic (slow) transaction.

spratt: Yeah...I think it's better to use locking here.  BLAH.

src/server/requestHandlers.js
```diff
@@ -97,24 +97,14 @@ exports.commentsOnLine = function commentsOnLine(request, response) {
...
-      multi.llen('comment:' + code_id + ':' + index);
-    });
-    multi.exec(function(err, lengths) {
-      var out = {};
-      lengths.forEach(function(length, i) {
-        out[indices[i]] = length;
-      });
-      return success(response, out);
-    });
+    return success(response, indices);
```
psimonyi, line 109: Now that this function is just "return that thing from the database", it would be nice to have a comment (probably for the whole function) saying that it returns a mapping from line number to number of comments. Otherwise you have to find out what the DB stores in `comment:<id>:indices` hashes to know what this function does.

psimonyi:
Does this fix #117?

src/server/requestHandlers.js
spratt, line 109: See new comment above this function.

Issue #123, opened 2013-04-04
Pass in ports from start scripts


psimonyi:
The various ports we use are set near the tops of a few JS files. The start scripts should set the ports and pass them to the JS, so they're configurable without changing the source.

psimonyi:
Closed by #126 since we now have /etc/server_devel.yml

PR #126, opened 2013-04-04
Modified startup script and server

bheesham:
- The server's configuration can now be set in `etc/server_{prod, devel}.yml`.
- The installation script installs redis.


etc/server_devel.yml
```diff
@@ -0,0 +1,4 @@
+server_port: 25414
+req_timeout: 3000
+
+# vim: ft=yaml
```
psimonyi, line 6: In modelines, I kind of prefer the full versions of commands (i.e. `filetype=yaml`), though I'd compromise to get this all on one line (especially since `ft` is a common enough option; more esoteric options really need to be readable).
It's probably a good idea to grab the standard vim modeline from the wiki, too.

src/server/index.js
```diff
@@ -7,14 +7,18 @@
...
 var connect = require('connect');
 var cors = require('cors');
 
 /******************************************************************************
 * Configuration                                                               *
 ******************************************************************************/
-var serverPort = 20193;
-var requestTimeout = 3000; // ms
+
+var server_config = require(process.env.SERVER_CONF);
```
psimonyi: Woah! You just _execute_ the YAML file?! Isn't there something like YAML.parse?

src/server/package.json
```diff
@@ -28,7 +28,8 @@
     "redis":            ">=0.8.2",
     "connect":          ">=2.7.3",
     "cors":             ">=0.0.5",
-    "proxy-middleware": ">=0.3.1"
+    "proxy-middleware": ">=0.3.1",
+		"js-yaml": 					">=2.0.3"
```
psimonyi: Spacing

bin/install_server.sh
```diff
@@ -14,6 +14,9 @@ ROOT_DIR=`pwd`
 ######################################################################
 # Configuration
 ######################################################################
+
+CLIENT_TARGET_DIR=~/webapps/cr_front
```
psimonyi: I don't like the idea of deploying to a home directory -- especially a hard-coded path.

bin/install_server.sh
```diff
@@ -22,16 +25,42 @@ NODE_VERSION=0.10.2
...
+tar -xf $REDIS_F.tar.gz -C $BIN_DIR/
+rm $REDIS_F.tar.gz
+mv -f $REDIS_F redis
+cd $BIN_DIR/redis 
+
+echo Compiling redis
+make
+
+echo Copying client files into the specified directory 
+cp -R $ROOT_DIR/src/client/* $CLIENT_TARGET_DIR/
```
psimonyi: It seems strange to be dealing with client files in `install_server.sh`. The file description at the top makes no mention of this.

bin/start.sh
```diff
@@ -18,7 +18,7 @@ NODE_BIN=$ROOT_DIR/bin/exe/node/bin
 NPM=$NODE_BIN/npm
 NODE=$NODE_BIN/node
 
-DB=redis-server
+DB=$ROOT_DIR/bin/exe/redis/src/redis-server
```
psimonyi: `bin/exe/redis/src` what?

bin/start.sh
```diff
@@ -62,7 +65,24 @@ ln -s -f $DB_LOG.$TIME.log $DB_LINK
...
+fi
+
+$NODE $INDEX_SCRIPT &> $SERVER_LOG.$TIME.log &
+
 echo $! > $SERVER_PID
 ln -s -f $SERVER_LOG.$TIME.log $SERVER_LINK
+
+
+
+
```
psimonyi: Extra blank lines at EOF

psimonyi:
It seems this still installs node, npm and redis even in development mode.

bin/install_server.sh
```diff
@@ -22,16 +25,42 @@ NODE_VERSION=0.10.2
...
 rm $NODE_F.tar.gz
-mv $NODE_F node
+mv -f $NODE_F node
+
+echo - complete!
+
+echo Installing component: redis
+
+echo - downloading...
+wget -q http://redis.googlecode.com/files/$REDIS_F.tar.gz
```
cdelahousse: Shouldn't we host a reliable version as we've done with our front end dependencies?

psimonyi: I thought we were hosting things only when we couldn't find a good download link.

bin/install_server.sh
spratt: Is there an obvious better way to do this?

bheesham:
@psimonyi , If you already have the things install, why install them again?
You should be using `start.sh`

``` bash
alias vim=sudo yum install vim && vim
```

bin/install_server.sh
spratt: I only hosted things when the only option was over https or there was no real download link.

spratt:
Outstanding issues for @bheesham:
- spacing in  `src/server/package.json`
- installing to a home directory in `bin/install_server.sh`
- add copying of client files to description of `bin/install_server.sh`

bheesham:
[Currently it does copy client files.](https://github.com/bheesham/Steamed-Pears/blob/master/bin/install_server.sh#L66)

spratt:
The problem is that there is no mention of copying client files in the description in the header of `bin/install_server.sh`

bheesham:
Ah, okay.

Also, for installation of the server to a home directory.
When you install this as root, and run it as another person, would that user not be able to find it? Because root's home dir is different.

spratt:
Why not reply directly to the comments on the lines of code?  It's easier to understand when there's context to what you're saying.

bheesham:
Because I'm replying directly to your comment.

spratt:
Okay, in general I agree with you.  So why are we installing to a home directory?

bin/install_server.sh
spratt: On this line, we install to a home directory.

spratt: Let's talk about this on this line.

psimonyi: I think the Right Thing To Do is have a good default like `/srv/code_review` or maybe `/var/www/code_review` and check an environment variable or command-line parameter for a user-specified value. (Think of installing programs, where the default is usually `/usr/bin` but you can do `./configure --prefix=$HOME/local` instead.) If you go for environment variables, look up Parameter Expansion in `bash(1)`; if you go for command-line parameters, look up getopt.

bin/install_server.sh
psimonyi: The main problem is not mentioning this in the description of what this file does, as @spratt says. A minor problem is that the file name `install_server.sh` suggests it's installing the server component, when it really seems to be installing everything.

psimonyi:
This will also fix #123.

bheesham:
Because the user we're installing it as, for sure has the permissions to write files there.
On WebFaction (our production server), that's the only place we're allowed to write.

psimonyi:
Let's discuss the home directory thing on its line.

bin/install_server.sh
psimonyi: @bheesham says we can be sure of write access to the home directory.
I think it is better to have a default for when it's your own server, and make it easy to override the default. (And you were worrying about what happens when installing as root before...)

bheesham: Hmm, yeah.
I guess when you run it as root, you'll probably want to put nginx, npm, and redis-server in `/usr/bin``.

As for not having the permissions to put it there, what would you suggest we do? Because changing PATH seems weird, and I feel gross. Because in development, I'd be running this a lot of times.

bin/install_server.sh
bheesham: So... do we not want to install everything?

bin/start.sh
```diff
@@ -1,23 +1,35 @@
...
 ######################################################################
 # Configuration
-NPM=npm
-NODE=node
 
-DB=redis-server
+if $PROD; then
+  NPM=$ROOT_DIR/bin/exe/node/bin/npm
+  NODE=$ROOT_DIR/bin/exe/node/bin/node
+  DB=$ROOT_DIR/bin/exe/redis/src/redis-server
```
psimonyi: These paths are no longer correct, right?

.gitignore
```diff
@@ -1,3 +1,5 @@
+*.swp
```
psimonyi: This should probably be `.*.swp` (dotfiles don't normally match `*` in globs). You've committed a swapfile in spite of this, anyway.

spratt: Ugh.  Let's get rid of the swapfile.

bin/install_server.sh
```diff
@@ -22,16 +22,61 @@ NODE_VERSION=0.10.2
...
+echo - extracting redis...
+tar -xf redis.tar.gz
+cd $TMP/$REDIS_F 
+
+echo - compiling redis...
+make
+
+echo - installing redis...
+cp $TMP/$REDIS_F/src/redis-server $BIN/redis-server
+cp $TMP/$REDIS_F/src/redis-cli $BIN/redis-cli
```
psimonyi: It would probably be less fragile to do `make PREFIX=$PREFIX install`.

bin/install_server.sh
```diff
@@ -22,16 +22,61 @@ NODE_VERSION=0.10.2
...
+# Redis version
+REDIS_VERSION=2.6.12
+
 ######################################################################
 
+TMP=$PREFIX/tmp
+BIN=$PREFIX/usr/bin
+
+mkdir -p $TMP 2> /dev/null
+mkdir -p $BIN 2> /dev/null
```
psimonyi: We probably want to know about any errors here. `mkdir -p` is silent when the directory exists, at least for me.

bin/install_server.sh
```diff
@@ -22,16 +22,61 @@ NODE_VERSION=0.10.2
 # Can be either 64 or 86
 NODE_ARCH=64
 
+# Redis version
+REDIS_VERSION=2.6.12
+
 ######################################################################
 
+TMP=$PREFIX/tmp
```
psimonyi: `TMP=${TEMP:-${TMP:-/tmp}}` would use the first non-empty path of `$TEMP`, `$TMP`, and `/tmp`. Any of those is probably a good temp directory. In any case, we should probably clean up after ourselves...

bin/install_server.sh
```diff
@@ -22,16 +22,61 @@ NODE_VERSION=0.10.2
...
+BIN=$PREFIX/usr/bin
```
psimonyi: Typically, `PREFIX` values are things like `/usr/local` or `$HOME` or `/usr`; in all cases, I'd expect binaries to go in `$PREFIX/bin`.

bheesham: Well, if it **is** /usr/, that sounds weird. /usr/node ? No thanks?

psimonyi: I'd still kind of like to see a default value for PREFIX, though I won't insist on it. However, the docs should say somewhere that you can use it (and generally tell you how to run this thing).

psimonyi: If `PREFIX` is `/usr`, then `BIN=$PREFIX/bin` would make `/usr/bin/node`, no?

psimonyi:
Blocking on the paths in `start.sh` and the swapfile; the other stuff can be done/discussed later if you want.

psimonyi:
You should also delete `bin/exe/.folder`.

psimonyi:
Perhaps `stop_dev.sh` should be renamed too. Does it work in production mode?

bin/install_server.sh
bheesham: Added.

bin/install_server.sh
bheesham: Added.

bin/install_server.sh
bheesham: Should have noted that I removed this after.

bin/install_server.sh
bheesham: Then we create /usr/tmp? Crazy talk.

bheesham:
It should now be mostly non-blocking. I didn't add `TMP=${...`, because I don't have permissions to write to compile in`/tmp`.

**THIS WORKS FOR ME IN PRODUCTION ON WEBFACTION.**

bin/install_server.sh
psimonyi: No, you put the temp files in `$TEMP` (or `$TMP`, or `/tmp` if neither is set), and install binaries to `$PREFIX/bin`.

bheesham: I can't compile in the `/tmp` folder. So, there is no installing binaries without compilation.

spratt:
Looks good.  I'm going to pull this down and try it out on my webfaction.

spratt:
Fantastic.  git gives errors pushing to webfaction if I don't have the docs submodule pulled in, but after installing it, it won't upload the docs to webfaction....fine, whatever git.  You suck.

bheesham:
Huh... I never got those errors.

spratt:
Redis fails to build:

```
make[3]: Entering directory `/home/sdp/tmp/redis-2.6.12/deps/jemalloc'
gcc -std=gnu99 -Wall -pipe -g3 -O3 -funroll-loops  -c -D_GNU_SOURCE -D_REENTRANT -Iinclude -Iinclude -o src/jemalloc.o src/jemalloc.c
gcc -std=gnu99 -Wall -pipe -g3 -O3 -funroll-loops  -c -D_GNU_SOURCE -D_REENTRANT -Iinclude -Iinclude -o src/arena.o src/arena.c
gcc -std=gnu99 -Wall -pipe -g3 -O3 -funroll-loops  -c -D_GNU_SOURCE -D_REENTRANT -Iinclude -Iinclude -o src/atomic.o src/atomic.c
gcc -std=gnu99 -Wall -pipe -g3 -O3 -funroll-loops  -c -D_GNU_SOURCE -D_REENTRANT -Iinclude -Iinclude -o src/base.o src/base.c
gcc -std=gnu99 -Wall -pipe -g3 -O3 -funroll-loops  -c -D_GNU_SOURCE -D_REENTRANT -Iinclude -Iinclude -o src/bitmap.o src/bitmap.c
gcc -std=gnu99 -Wall -pipe -g3 -O3 -funroll-loops  -c -D_GNU_SOURCE -D_REENTRANT -Iinclude -Iinclude -o src/chunk.o src/chunk.c
gcc -std=gnu99 -Wall -pipe -g3 -O3 -funroll-loops  -c -D_GNU_SOURCE -D_REENTRANT -Iinclude -Iinclude -o src/chunk_dss.o src/chunk_dss.c
gcc -std=gnu99 -Wall -pipe -g3 -O3 -funroll-loops  -c -D_GNU_SOURCE -D_REENTRANT -Iinclude -Iinclude -o src/chunk_mmap.o src/chunk_mmap.c
gcc -std=gnu99 -Wall -pipe -g3 -O3 -funroll-loops  -c -D_GNU_SOURCE -D_REENTRANT -Iinclude -Iinclude -o src/ckh.o src/ckh.c
gcc -std=gnu99 -Wall -pipe -g3 -O3 -funroll-loops  -c -D_GNU_SOURCE -D_REENTRANT -Iinclude -Iinclude -o src/ctl.o src/ctl.c
gcc -std=gnu99 -Wall -pipe -g3 -O3 -funroll-loops  -c -D_GNU_SOURCE -D_REENTRANT -Iinclude -Iinclude -o src/extent.o src/extent.c
gcc -std=gnu99 -Wall -pipe -g3 -O3 -funroll-loops  -c -D_GNU_SOURCE -D_REENTRANT -Iinclude -Iinclude -o src/hash.o src/hash.c
gcc -std=gnu99 -Wall -pipe -g3 -O3 -funroll-loops  -c -D_GNU_SOURCE -D_REENTRANT -Iinclude -Iinclude -o src/huge.o src/huge.c
gcc -std=gnu99 -Wall -pipe -g3 -O3 -funroll-loops  -c -D_GNU_SOURCE -D_REENTRANT -Iinclude -Iinclude -o src/mb.o src/mb.c
gcc -std=gnu99 -Wall -pipe -g3 -O3 -funroll-loops  -c -D_GNU_SOURCE -D_REENTRANT -Iinclude -Iinclude -o src/mutex.o src/mutex.c
gcc -std=gnu99 -Wall -pipe -g3 -O3 -funroll-loops  -c -D_GNU_SOURCE -D_REENTRANT -Iinclude -Iinclude -o src/prof.o src/prof.c
gcc -std=gnu99 -Wall -pipe -g3 -O3 -funroll-loops  -c -D_GNU_SOURCE -D_REENTRANT -Iinclude -Iinclude -o src/quarantine.o src/quarantine.c
gcc -std=gnu99 -Wall -pipe -g3 -O3 -funroll-loops  -c -D_GNU_SOURCE -D_REENTRANT -Iinclude -Iinclude -o src/rtree.o src/rtree.c
gcc -std=gnu99 -Wall -pipe -g3 -O3 -funroll-loops  -c -D_GNU_SOURCE -D_REENTRANT -Iinclude -Iinclude -o src/stats.o src/stats.c
gcc -std=gnu99 -Wall -pipe -g3 -O3 -funroll-loops  -c -D_GNU_SOURCE -D_REENTRANT -Iinclude -Iinclude -o src/tcache.o src/tcache.c
gcc -std=gnu99 -Wall -pipe -g3 -O3 -funroll-loops  -c -D_GNU_SOURCE -D_REENTRANT -Iinclude -Iinclude -o src/util.o src/util.c
gcc -std=gnu99 -Wall -pipe -g3 -O3 -funroll-loops  -c -D_GNU_SOURCE -D_REENTRANT -Iinclude -Iinclude -o src/tsd.o src/tsd.c
ar crus lib/libjemalloc.a src/jemalloc.o src/arena.o src/atomic.o src/base.o src/bitmap.o src/chunk.o src/chunk_dss.o src/chunk_mmap.o src/ckh.o src/ctl.o src/extent.o src/hash.o src/huge.o src/mb.o src/mutex.o src/prof.o src/quarantine.o src/rtree.o src/stats.o src/tcache.o src/util.o src/tsd.o
make[3]: Leaving directory `/home/sdp/tmp/redis-2.6.12/deps/jemalloc'
make[2]: Leaving directory `/home/sdp/tmp/redis-2.6.12/deps'
    CC adlist.o
    CC ae.o
    CC anet.o
    CC dict.o
    CC redis.o
    CC sds.o
    CC zmalloc.o
    CC lzf_c.o
    CC lzf_d.o
    CC pqsort.o
    CC zipmap.o
    CC sha1.o
    CC ziplist.o
    CC release.o
    CC networking.o
    CC util.o
    CC object.o
    CC db.o
    CC replication.o
    CC rdb.o
    CC t_string.o
    CC t_list.o
    CC t_set.o
    CC t_zset.o
    CC t_hash.o
    CC config.o
    CC aof.o
    CC pubsub.o
    CC multi.o
    CC debug.o
    CC sort.o
    CC intset.o
    CC syncio.o
    CC migrate.o
    CC endianconv.o
    CC slowlog.o
    CC scripting.o
    CC bio.o
    CC rio.o
    CC rand.o
    CC memtest.o
    CC crc64.o
    CC bitops.o
    CC sentinel.o
    LINK redis-server
zmalloc.o: In function `zmalloc_used_memory':
/home/sdp/tmp/redis-2.6.12/src/zmalloc.c:223: undefined reference to `__sync_add_and_fetch_4'
collect2: ld returned 1 exit status
make[1]: *** [redis-server] Error 1
make[1]: Leaving directory `/home/sdp/tmp/redis-2.6.12/src'
make: *** [all] Error 2
cd src && make install
make[1]: Entering directory `/home/sdp/tmp/redis-2.6.12/src'
    LINK redis-server
zmalloc.o: In function `zmalloc_used_memory':
/home/sdp/tmp/redis-2.6.12/src/zmalloc.c:223: undefined reference to `__sync_add_and_fetch_4'
collect2: ld returned 1 exit status
make[1]: *** [redis-server] Error 1
make[1]: Leaving directory `/home/sdp/tmp/redis-2.6.12/src'
make: *** [install] Error 2
- installing redis...
cp: cannot stat `/home/sdp/tmp/redis-2.6.12/src/redis-server': No such file or directory
cp: cannot stat `/home/sdp/tmp/redis-2.6.12/src/redis-cli': No such file or directory
```

bin/install_server.sh
```diff
@@ -22,16 +22,69 @@ NODE_VERSION=0.10.2
...
+
+echo - extracting redis...
+tar -xf redis.tar.gz
+rm redis.tar.gz
+
+cd $TMP/$REDIS_F 
+
+echo - compiling redis...
+make
+make PREFIX=$PREFIX install
```
spratt: I think this should be `PREFIX=$PREFIX make install`

spratt:
The install script works fine on my mac using clang 4.2, but it won't compile on my webfaction host using gcc 4.1.  I have a suspicion that gcc 4.1 is a little out of date.

spratt:
Bhee, I fixed the issues I mentioned above in a pull request to your master branch.

spratt:
[blocking] I can't execute the node binary file that we download using this script.

spratt:
For the record, I found the solution to the redis compile bug from: http://www.eschrade.com/page/undefined-reference-to-__sync_add_and_fetch_4/

spratt:
`uname -m` returns an architecture that my mac's version of gcc doesn't support...and the redis build uses gcc in particular at one stage of its build (as opposed to $CC which would use clang).

spratt:
This now works on the main site.  See: http://review.steamedpears.com

spratt:
@psimonyi you should take a look at this request now.  I think a lot of issues have been fixed.

bin/install_server.sh
```diff
@@ -14,24 +15,80 @@ ROOT_DIR=`pwd`
 ######################################################################
 # Configuration
 ######################################################################
-BIN_DIR=$ROOT_DIR/bin/exe
 
 # A valid node version
-NODE_VERSION=0.10.2
+NODE_VERSION=0.10.4
```
psimonyi, line 20: The "Requirements" section should probably list this version of Node.

bin/install_server.sh
```diff
@@ -14,24 +15,80 @@ ROOT_DIR=`pwd`
...
 
 # Can be either 64 or 86
 NODE_ARCH=64
 
+# Redis version
+REDIS_VERSION=2.6.12
```
psimonyi, line 26: Same here. The docs say we've tested with 2.6.10.

bin/install_server.sh
```diff
@@ -14,24 +15,80 @@ ROOT_DIR=`pwd`
...
+wget -O redis.tar.gz -q http://redis.googlecode.com/files/$REDIS_F.tar.gz
+
+echo - extracting redis...
+tar -xf redis.tar.gz
+rm redis.tar.gz
+
+cd $TMP/$REDIS_F 
+
+echo - compiling redis...
+CFLAGS=-march=`uname -m` make
```
psimonyi: Waiting on this thing.

bin/build_production_all.sh
```diff
@@ -13,14 +13,12 @@ CLIENT_SCRIPTS_DIR="$CLIENT_DIR/scripts"
...
-  #Defaults to `production/`
-  TARGET="$PWD/production"
-fi
+# Defaults to `production/`
+TARGET=${TARGET:-$PWD/production}
 
 #Temporary folder for the build
-TARGET_TMP="$TARGET""_tmp"
+TMP=${TEMP:-${TMP:-/tmp}}
+TARGET_TMP=$TMP/cr
```
psimonyi: It would be better to use mktemp for this (`mktemp -d --tmpdir="$TMP" codereview-XXXX`). In fact, just `--tmpdir` (no argument) would use `$TMPDIR` if set, and default to `/tmp`.
This can wait for later. [non-blocking]

bin/install_server.sh
spratt: By which, Peter means that although it fixes an issue on the server it breaks the build everywhere else.

psimonyi: Right. I think the appropriate thing to do is just call `make` here, and do `export CFLAGS=i686` before running the script, because this seems like a bug caused by an old version of gcc. Other possibilities, since Webfaction is currently our primary deployment target, are: have a script `webfaction_hack.bash` that does that, or try to detect the version of GCC (but that's not exactly easy).

spratt: I like the export CFLAGS fix.  We can document the issue and the resolution.

spratt:
This pull request can't be automatically merged...

bheesham:
It can now.

spratt:
Builds nicely on my mac.  Fails on the server without -march=i686.  Now testing with -march=i686.

spratt:
Builds successfully on server.  I'll do one last review then merge.

bin/install_server.sh
spratt, line 20: Fixed.

bin/install_server.sh
spratt, line 26: Fixed.

PR #137, opened 2013-04-10
Fixed Submodule

cdelahousse:
It works if you try 
`git clone --recurse-submodules git@github.com:cdelahousse/Steamed-Pears.git CodeReview`


psimonyi:
So basically, it shouldn't have been in gitignore? That kind of makes sense, actually.

PR #138, opened 2013-04-10
More gitignore cleanup

psimonyi:
I missed this the first time.

### DIFF
--- a/src/server/.gitignore
+++ b/src/server/.gitignore
@@ -1,8 +1,0 @@
-*.pid
-db_info.js
-node_modules
-models/config/
-models/migrations/
-node_modules*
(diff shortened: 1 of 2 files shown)


src/server/.gitignore
```diff
@@ -1,8 +0,0 @@
-*.pid
-db_info.js
-node_modules
```
cdelahousse, line 3: Don't we need this?

psimonyi, line 3: Yes, but it's covered by the addition to the `.gitignore` in the root. (Incidentally, that will ignore any `node_modules` directory in the project, which is probably what we want).

PR #146, opened 2013-04-11
bookmarklet link dynamically generated

cdelahousse:


psimonyi:
Uh, are you missing a file? I don't see the code to generate the link anywhere.

psimonyi:
... or I'm blind, or the page hadn't finished loading.

psimonyi:
Should this be an inline script? Maybe it would be cleaner as a separate file. I'm not sure; does anyone else have an opinion?

cdelahousse:
I figured if was nicer to keep it all together.

All the script files were nice and encapsulated and I didn't want to add another $(document).ready() function.

psimonyi:
Okay, sure. Seems fine for now, anyway.

PR #149, opened 2013-04-12
windows.location.origin is not supported by FF

cdelahousse:
Fucking Simon and the hilarity he sees in my honest observations!!!!

### DIFF
--- a/src/client/index.html
+++ b/src/client/index.html
@@ -18,10 +18,9 @@
       by dragging the link to your bookmarks bar.
       <script>
         (function() {
-
           var bookmarklet_link =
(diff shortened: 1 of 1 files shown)


psimonyi:
Firefox 21 supports it.
https://developer.mozilla.org/en-US/docs/DOM/window.location

spratt:
@psimonyi We discovered that.  But that it is supported in the next version of FF is not a good reason to use it.

bheesham:
lol

spratt:
Screen Shot 2013-04-12 at 3 04 52 PM

PR #160, opened 2013-04-19
Bookmarklet bugfixes

cdelahousse:
- Now we can serve to pages that SSL
- Copies over style.css to production
- Updated bookmarklet link

### DIFF
--- a/src/bookmarklet/Makefile
+++ b/src/bookmarklet/Makefile
@@ -33,6 +33,7 @@ build-local: buildvendors
 .PHONY: build
 build: build-local make-target-dirs move-to-target
 	cp bookmarklet.js $(TARGET)
+	cp style.css $(TARGET)
 
 # Build the production version of the bookmarklet to $TARGET
 .PHONY: build-production
@@ -48,6 +49,9 @@ build-production: make-target-dirs buildvendors
 		echo "Minifying vendors/$$VDR"; \
 		uglifyjs -o "$(TARGET)/$(VENDORS)/$$VDR" "$(VENDORS)/$$VDR"; \
 	done
+	@echo "Copy over the CSS"
+	cp style.css $(TARGET)
+
 
 #Reset bookmarklet by deleting all generated files and redownloading dependencies
 .PHONY: reset
--- a/src/bookmarklet/bookmarklet.js
+++ b/src/bookmarklet/bookmarklet.js
@@ -22,7 +22,7 @@ g.vendorUri = '/bookmarklet/vendors/';
 g.vendorDefault = 'readability';
 g.bookmarkletUri =  '/bookmarklet';
 g.newCodeApi = '/do/newcode';
-g.baseUrl = g.debug ? '//localhost:8080' : '//review.steamedpears.com';
+g.baseUrl = g.debug ? 'http://localhost:8080' : 'http://review.steamedpears.com';
 
 function dbg() { if (g.debug) console.log.apply(console,arguments); }
 
--- a/src/client/index.html
+++ b/src/client/index.html
@@ -21,7 +21,7 @@
           var bookmarklet_link =
           "javascript:(function(){window.codeReview&&window.codeReview.exit();var a=document.createElement('script');a.src='"
             + window.location.protocol + '//' + window.location.host
-            + "/bookmarklet/bookmarklet.js';(document.head||document.body).appendChild(a)})();";
+            + "/bookmarklet/bookmarklet.min.js';(document.head||document.body).appendChild(a)})();";
           document.getElementById("bookmarklet").href = bookmarklet_link;
         })();
       </script>


cdelahousse:
ping @spratt

src/client/index.html
```diff
@@ -21,7 +21,7 @@ <h1 id="title"><a href="index.html">CodeReview</a></h1>
           var bookmarklet_link =
           "javascript:(function(){window.codeReview&&window.codeReview.exit();var a=document.createElement('script');a.src='"
             + window.location.protocol + '//' + window.location.host
-            + "/bookmarklet/bookmarklet.js';(document.head||document.body).appendChild(a)})();";
+            + "/bookmarklet/bookmarklet.min.js';(document.head||document.body).appendChild(a)})();";
```
spratt, line 24: Does this work even in a non-production build?  Is bookmarklet.min.js the canonical form?  Have you tested this?

cdelahousse, line 24: I tested it by moving running my server locally, dragging the bookmarklet over and editing it a bit. I changed localhost:8080 to review.steamedpears.com. The part that this pull request changes was untouched by the testing.

It works that way.

spratt, line 24: Does it work if you build the bookmarklet in dev mode?  I think the root cause of this issue was that dev mode builds bookmarklet.js and production builds bookmarklet.min.js

cdelahousse, line 24: Oh. No. bookmarklet.min.js is for the minified/production version, bookmarklet.js is for the dev/unminified version.

I think there should be a distinction between the min and non-min versions. That's why I had so many links to the bookmarklet on my test page.

spratt, line 24: ...so this line is broken in a dev environment?

cdelahousse, line 24: Yes, which doesn't matter because I have a bookmarklet test page.

spratt, line 24: It would be nice if we could get this to work in both environments.  Why not always generate and link to bookmarklet.js, but it will be minified in production and regular in development?

psimonyi, line 24: I like the sound of that idea.

src/bookmarklet/Makefile
```diff
@@ -33,6 +33,7 @@ build-local: buildvendors
 .PHONY: build
 build: build-local make-target-dirs move-to-target
 	cp bookmarklet.js $(TARGET)
+	cp style.css $(TARGET)
```
psimonyi, line 36: Why isn't this done by the `move-to-target` rule?

src/bookmarklet/Makefile
```diff
@@ -48,6 +49,9 @@ build-production: make-target-dirs buildvendors
 		echo "Minifying vendors/$$VDR"; \
 		uglifyjs -o "$(TARGET)/$(VENDORS)/$$VDR" "$(VENDORS)/$$VDR"; \
 	done
+	@echo "Copy over the CSS"
+	cp style.css $(TARGET)
```
psimonyi, line 53: a) The other echos here use a gerund: "Copying over the CSS"
b) The `cp` is the only command not silenced by `@`; why?
c) What information does "Copy over the CSS" convey that the actual `cp` command doesn't? What does it add?
d) Why do you need to copy style.css here and in the `build` target? And in `move-to-target`?
I'm not really a fan of the silencing in this makefile, so I'd just deleting line 52 if you still need to do the copying here.

src/bookmarklet/Makefile
```diff
@@ -48,6 +49,9 @@ build-production: make-target-dirs buildvendors
...
+
```
psimonyi, line 54: Extra blank line.

src/bookmarklet/bookmarklet.js
```diff
@@ -22,7 +22,7 @@ g.vendorUri = '/bookmarklet/vendors/';
 g.vendorDefault = 'readability';
 g.bookmarkletUri =  '/bookmarklet';
 g.newCodeApi = '/do/newcode';
-g.baseUrl = g.debug ? '//localhost:8080' : '//review.steamedpears.com';
+g.baseUrl = g.debug ? 'http://localhost:8080' : 'http://review.steamedpears.com';
```
bheesham, line 25: If you just do `://` here, will the browser put in which protocol? I know browsers do this for resources located on the page, but not sure about the bookmarks bar.

It'd be cool if our user was using SSL, then we should use SSL too. Or just always use SSL?

psimonyi, line 25: Yeah, except review.steamedpears.com doesn't do SSL right now.

bheesham, line 25: Hmm. We'll have to work on that. I think we can get a free certificate 
from Webfaction, but it'll always give a warning because the domain's 
don't match up.

Something to think about for the future.

psimonyi, line 25: A not-quite-valid cert is not really any more useful than no cert, except in testing. I don't think the user would even be prompted, because it's just a script that has the error, not the page itself.

psimonyi, line 25: Also, this will not work starting in Firefox 23, when mixed active content blocking is turned on by default. I'm not sure what other browsers do about http scripts on https pages.

spratt:
Good enough.

